PR #3, opened 2017-09-13
Fix #1. Workaround for angular not cleaning up routing events when module is destroyed.

### LINKED ISSUE #1: angular2 router not deactivated
Hi,
first of all great work!
I'm trying to use your example to see if this tool fit with my requirement and in the angular2 application I tried to add a simple router example. the angular 2 works great but when I try to switch to react (after angular2 has been loaded) it fail because angular2 is not unmounted and it try to find the path app1.

this is the repository that I created: https://github.com/daniele-zurico/microservice-test

joeldenning:
See #1, https://github.com/CanopyTax/single-spa/issues/115, https://github.com/angular/angular/issues/19079, and http://plnkr.co/edit/43a4sWEsdm5yfpVkJ5Zs?p=preview

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,19 +5,20 @@ Helpers for building [single-spa](https://github.com/CanopyTax/single-spa) appli
 An example can be found in the [single-spa-examples](https://github.com/CanopyTax/single-spa-examples/blob/master/src/angular2/angular2.app.js) repository.
 
 ## Quickstart
-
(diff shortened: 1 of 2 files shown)


joeldenning:
@daniele-zurico could you review this, since you were the one who originally found this workaround?

daniele-zurico:
I think that is missed also the import of the Router? Or maybe I missed something

joeldenning:
ah yes, you're right about both of those things. I was lazy in creating this, should have been a bit more thorough

joeldenning:
Just pushed up the fixes and some documentation. Also tested it with your microservice-test repository and it worked!

daniele-zurico:
It's perfect now!

joeldenning:
Released as [v1.1.0](https://github.com/CanopyTax/single-spa-angular2/releases/tag/v1.1.0) to npm

PR #8, opened 2018-03-14
Adding ability to pass domElementGetter as a custom prop.

joeldenning:


src/single-spa-angular2.js
```diff
@@ -45,18 +41,29 @@ function bootstrap(opts) {
...
+  return Promise
+    .resolve()
+    .then(() => {
+      const domElementGetter = chooseDomElementGetter(opts, props)
+      if (!domElementGetter) {
+        throw new Error(`cannot mount angular application '${props.name || props.appName}' without a domElementGetter provided either as an opt or a prop`)
+      }
+
+      const containerEl = getContainerEl(domElementGetter)
+      containerEl.innerHTML = opts.template
```
frehner: do you need to be returning something in this promise for the next `.then(domElementGetter => ...` to work?

hm, maybe not, since you're not actually using domElementGetter in the next .then. so maybe just remove it so you're good?

joeldenning: yeah I thought i was going to use domElementGetter but I'm not. Will change 👍

PR #13, opened 2018-12-29
Support for angular 7, including angular@router 7.

joeldenning:
There was some change with Angular 7 that caused us to need to do an applicationRef.tick() whenever a popstate or hashchange event goes through single-spa. See the comment in the code in this PR for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,16 +18,16 @@ import mainModule from './main-module.ts';
 import {Router} from '@angular/router';
 
 export default singleSpaAngular({
-	domElementGetter,
-	mainModule,
-	angularPlatform: platformBrowserDynamic(),
-	template: `<component-to-render />`,
-	Router,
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -12,30 +12,20 @@ First, in the [single-spa application](https://github.com/CanopyTax/single-spa/b
 
 ```js
 import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
+import {ApplicationRef} from '@angular/core';
 import singleSpaAngular from 'single-spa-angular';
 import mainModule from './main-module.ts';
 import {Router} from '@angular/router';
 
-const ngLifecycles = singleSpaAngular({
+export default singleSpaAngular({
```
joeldenning: I'm working on some things that will end up making `export default` be the right thing to do.

PR #20, opened 2019-03-16
remove babel

cgatian:
Note: also included tslint in this PR


package.json
```diff
@@ -33,11 +34,9 @@
     "@angular-devkit/architect": "^0.13.6",
     "@angular-devkit/build-angular": "^0.13.6",
     "@angular-devkit/core": "^7.3.6",
-    "@babel/cli": "^7.2.3",
-    "@babel/core": "^7.2.2",
-    "@babel/preset-env": "^7.2.3",
     "@types/node": "^11.11.3",
     "rimraf": "^2.6.2",
+    "tslint": "^5.14.0",
```
joeldenning: My understanding is that tslint is being deprecated in favor of typescript support in eslint. See https://medium.com/palantir/tslint-in-2019-1a144c2317a9. Should we use eslint instead of tslint?

joeldenning:
We should remove the `.babelrc` file as part of this PR

package.json
cgatian: I saw this as well but hought that was the general announcement and it wasn't ready?

> Once we consider ESLint feature-complete w.r.t. TSLint, we will deprecate TSLint

PR #22, opened 2019-03-16
feat(builder): Adding ZoneJS to externals

joeldenning:


src/builders/browser/index.ts
```diff
@@ -40,13 +41,15 @@ export class SingleSpaBrowserBuilder extends BrowserBuilder {
...
     }
 
-    // Configure Wepack output
-    config.output = {
-      library: options.libraryName,
-      libraryTarget: options.libraryTarget
-    };
-
-    return config;
+    return webpackMerge.smart(config, {
```
joeldenning: See https://github.com/survivejs/webpack-merge#smart-merging.

The previous code overwrote the entire output object, but now we just merge in the library and libraryTarget. See https://webpack.js.org/configuration/output/#root for more details

src/builders/browser/index.ts
```diff
@@ -40,13 +41,15 @@ export class SingleSpaBrowserBuilder extends BrowserBuilder {
...
+      output: {
+        library: options.libraryName,
+        libraryTarget: options.libraryTarget,
+      },
+      externals: {
+        'zone.js': 'Zone',
```
joeldenning: all imports of `zone.js` will use a global variable called `Zone`

src/builders/browser/index.ts
cgatian: Cool! 👍

PR #25, opened 2019-03-16
feat(builder): use project name if libraryName is not configured

### LINKED ISSUE #17: Use project name as default
The builder should default to the angular cli project name for the libraryName.

cgatian:
Closes #17

### DIFF
--- a/src/builders/browser/index.ts
+++ b/src/builders/browser/index.ts
@@ -1,6 +1,7 @@
+import { BuilderContext } from '@angular-devkit/architect';
 import { BrowserBuilder } from '@angular-devkit/build-angular';
 import { Path, virtualFs } from '@angular-devkit/core';
-import { BuilderContext } from '@angular-devkit/architect';
+import { camelize } from '@angular-devkit/core/src/utils/strings';
 import * as fs from 'fs';
 import { Configuration } from 'webpack';
 import webpackMerge from 'webpack-merge';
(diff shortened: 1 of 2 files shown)


cgatian:
I just thought of a scenario this doesn't work with. Reopening a new pr

src/builders/browser/index.ts
```diff
@@ -16,8 +17,15 @@ export class SingleSpaBrowserBuilder extends BrowserBuilder {
...
     projectRoot: Path,
     host: virtualFs.Host<fs.Stats>,
-    options: SingleSpaBuilderSchema
+    options: SingleSpaBuilderSchema,
   ): Configuration {
+
+    const libraryName = options.libraryName || this.getProjectName();
+
+    if (!libraryName) {
+      throw new Error('Unable to determine libraryName.');
```
joeldenning: In webpack, `output.library` is not required for all possible values of `output.libraryTarget`. Specifically, the `system` libraryTarget does not require a name, and actually SystemJS@3 does not support giving it one. So I think we should remove this error check. Instead we should let webpack do it's validation on the webpack config, which will include throwing an error if the libraryName is needed but not defined.

src/builders/browser/index.ts
```diff
@@ -43,13 +51,19 @@ export class SingleSpaBrowserBuilder extends BrowserBuilder {
...
       },
       externals: {
         'zone.js': 'Zone',
       },
-    })
+    });
+  }
+
+  private getProjectName() {
+    if (this.context.targetSpecifier && this.context.targetSpecifier.project) {
```
joeldenning: 👍 I like using this as a default.

src/builders/browser/index.ts
```diff
@@ -43,13 +51,19 @@ export class SingleSpaBrowserBuilder extends BrowserBuilder {
...
+      return camelize(this.context.targetSpecifier.project);
```
joeldenning: camelize is nice when the libraryName is a global variable in the DOM, but maybe confusing or not nice when it's not. For example, if you have a project called `angular-app`, it would be confusing if it went into SystemJS as `angularApp` instead of `angular-app`. I think I would prefer just leaving the project name alone for this reason, and letting people specify a custom name if they want one.

cgatian:
Fixed conflicts. Removed throwing exception

PR #34, opened 2019-03-22
Added support for ng serve

pauljmelia:
Extended DevServerBuilder from @angular-devkit/build-angular and overrode buildWebpackConfig to call SingleSpaBrowserBuilder.buildWepackConfig by @cgatian


cgatian:
Awesome! Ill take a look tonight!!! Thank you

.babelrc
```diff
@@ -0,0 +1,5 @@
+{
+  "presets": [
+    "@babel/preset-env",
```
joeldenning: We just recently moved away from babel in favor of just typescript. See related https://github.com/CanopyTax/single-spa-angular/issues/16.

It looks like this `.babelrc` file isn't being used -- let's remove it

src/builders/browser/schema.d.ts
```diff
@@ -9,7 +9,7 @@
 import { NormalizedBrowserBuilderSchema } from '@angular-devkit/build-angular';
 
 export interface SingleSpaBuilderSchema extends NormalizedBrowserBuilderSchema {
-  libraryTarget: string;
+  libraryTarget: "var" | "assign" | "this" | "window" | "global" | "commonjs" | "commonjs2" | "amd" | "umd" | "jsonp" | undefined;
```
joeldenning: This is a duplication of [this type in the webpack repo](https://github.com/webpack/webpack/blob/0c1beb99a82bdd8d262673b7c54d805135ffd91a/declarations/WebpackOptions.d.ts#L1111). I'm not an expert in typescript -- is there a way to use that definition instead of duplicate it?

See related https://github.com/webpack/webpack/pull/8864 which is going to add one more possible libraryTarget -- just an example of where it would be better to reuse/inherit webpack's definitive definition instead of duplicating our own.

src/builders/dev-server/index.ts
```diff
@@ -0,0 +1,29 @@
...
+import { BuilderContext } from '@angular-devkit/architect';
+import { virtualFs, Path, tags} from '@angular-devkit/core';
+import { SingleSpaBuilderSchema } from '../browser/schema';
+import { Configuration } from 'webpack';
+import * as fs from 'fs';
+import { SingleSpaBrowserBuilder } from '../browser';
+
+
+export class SingleSpaDevServer extends DevServerBuilder {
+    constructor(context: BuilderContext) {
```
joeldenning: nit: this constructor is unnecessary, since all we're doing is calling `super`, which is what happens by default if you omit a constructor. So we could remove this entirely

joeldenning:
Additionally, we should make this become the default `ng serve` builder. See https://github.com/CanopyTax/single-spa-angular/blob/master/src/schematics/ng-add/index.ts#L72 for how we're doing that with `ng-build`.

joeldenning:
One more thing - we will want to add a new entry to `scripts` for running the single-spa webpack dev server manually. With `ng build`, we add the `build:single-spa` script to the package.json for this.

For dev server, let's call it `serve:single-spa`. See https://github.com/CanopyTax/single-spa-angular/blob/master/src/schematics/ng-add/index.ts#L94 for the pattern to follow

joeldenning:
Oops and one last thing -- I have added instructions on how to test all the schematics stuff locally:

https://github.com/CanopyTax/single-spa-angular/blob/master/CONTRIBUTING.md

pauljmelia:
Thanks @joeldenning I'll try and take a look over the weekend

.babelrc
pauljmelia: My bad - or rather SourceTree's bad: seems like "use theirs" conflict resolution gets confused when "theirs" no longer contains the file.

src/builders/dev-server/index.ts
pauljmelia: sure: I was just following the precedent set in SingleSpaBrowserBuilder for consitency.

src/builders/browser/schema.d.ts
pauljmelia: I have to admit, I'm no typescript expert either. From my understanding though, the typescript here is generated from schema.json. 

As an experiment, I changed the definition in the schema to use a $ref to the definition in webpack/schemas/WebpackOptions.json and then used [json-schema-to-typescript](https://www.npmjs.com/package/json-schema-to-typescript) to generate a new schema file: the output was the same as I committed. 

I'm not happy with the fact that the $ref was pointing to ~/../../../node_modules/webpack/... so I won't be adding that to the PR.

pauljmelia:
> Additionally, we should make this become the default `ng serve` builder. See https://github.com/CanopyTax/single-spa-angular/blob/master/src/schematics/ng-add/index.ts#L72 for how we're doing that with `ng-build`.

I see that you're adding a new node underneath `architect` for `single-spa`, for the serve declaration, this would have to add another node under architect (`single-spa-serve`?), would you want to suffix the build node with `-build` for clarity or leave it as is?

src/builders/dev-server/index.ts
pauljmelia: To keep consistency with what's in the repo already I've left this in for now.

joeldenning: fair enough 👍

src/builders/browser/schema.d.ts
joeldenning: cool we can just update this when that webpack PR gets merged in

pauljmelia:
Thanks @joeldenning, very happy to contribute!

PR #38, opened 2019-03-26
 Having ng serve load up the single-spa html file. Resolves #37. 

### LINKED ISSUE #37: Allow user to configure where their single-spa root config is during ng serve
In order for single-spa-angular's dev server to be useful, we need it to load up the [single-spa root config](https://github.com/CanopyTax/single-spa-angular#check-if-it-works) as the [webpack-dev-server's `index`](https://webpack.js.org/configuration/dev-server/#devserverindex) or via the [webppack-dev-serve`openPage`](https://webpack.js.org/configuration/dev-server/#devserveropenpage) configuration options. `index` would be for a static index.html file (probably most common for single-spa users), or `openPage` would be if the index.html is hosted by a local server.

I think the ideal solution would be one where we add a question during the `ng add single-spa-angular` process that let's the user choose between the two options. This would add some options to the `single-spa-angular:dev-server` schematic that could be modified via the angular.json file.

joeldenning:
See #37

### DIFF
--- a/src/builders/dev-server/index.ts
+++ b/src/builders/dev-server/index.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import { DevServerBuilder } from '@angular-devkit/build-angular';
 import { BuilderContext } from '@angular-devkit/architect';
 import { virtualFs, Path, tags } from '@angular-devkit/core';
--- a/src/builders/dev-server/index.ts
+++ b/src/builders/dev-server/index.ts
@@ -1,12 +1,15 @@
+// @ts-nocheck
 import { DevServerBuilder } from '@angular-devkit/build-angular';
 import { BuilderContext } from '@angular-devkit/architect';
-import { virtualFs, Path, tags} from '@angular-devkit/core';
+import { virtualFs, Path, tags } from '@angular-devkit/core';
 import { SingleSpaBuilderSchema } from '../browser/schema';
 import { Configuration } from 'webpack';
 import * as fs from 'fs';
+import * as path from 'path';
 import { SingleSpaBrowserBuilder } from '../browser';
+import * as webpackMerge from 'webpack-merge';
 
-
+// @ts-ignore
 export class SingleSpaDevServer extends DevServerBuilder {
     constructor(context: BuilderContext) {
         super(context);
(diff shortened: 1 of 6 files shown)


src/builders/dev-server/index.ts
```diff
@@ -1,12 +1,14 @@
...
+import { virtualFs, Path, tags } from '@angular-devkit/core';
 import { SingleSpaBuilderSchema } from '../browser/schema';
 import { Configuration } from 'webpack';
 import * as fs from 'fs';
+import * as path from 'path';
 import { SingleSpaBrowserBuilder } from '../browser';
+import * as webpackMerge from 'webpack-merge';
 
-
+// @ts-ignore
```
joeldenning: The `_buildServerConfig` function is the only way to override angular's default dev server configuration. But typescript freaks out because it's a private method, so I disabled that check.

src/builders/dev-server/index.ts
```diff
@@ -24,6 +26,17 @@ export class SingleSpaDevServer extends DevServerBuilder {
         SINGLE-SPA-ANGULAR: Angular dev server is serving application as a single module`);
         return webpackConfig;
     }    
+    _buildServerConfig(root, projectRoot, options, browserOptions) {
+        // @ts-ignore
+        const devServerConfig = super._buildServerConfig(root, projectRoot, options, browserOptions);
```
joeldenning: typescript not liking me calling a private method

src/builders/dev-server/index.ts
```diff
@@ -24,6 +26,17 @@ export class SingleSpaDevServer extends DevServerBuilder {
...
+
+        return webpackMerge.smart(devServerConfig, {
+            // @ts-ignore
```
joeldenning: the typescript definition that is used by webpack-merge doesn't know that `contentBase` is a thing, but [it is](https://webpack.js.org/configuration/dev-server/#devservercontentbase)

src/builders/dev-server/index.ts
```diff
@@ -24,6 +26,17 @@ export class SingleSpaDevServer extends DevServerBuilder {
...
+            contentBase: path.resolve(root, projectRoot.serveDirectory || '../'),
```
joeldenning: A directory that will serve static assets, including the [single-spa root config's html file](https://github.com/CanopyTax/single-spa-angular#check-if-it-works)

src/builders/dev-server/index.ts
```diff
@@ -24,6 +26,17 @@ export class SingleSpaDevServer extends DevServerBuilder {
...
+            historyApiFallback: true,
+            publicPath: root.slice(root.lastIndexOf(path.sep)) + '/' + options.outputPath,
```
joeldenning: example: `/test1/dist/test1/`

This is where the webpack assets are served from.

src/builders/dev-server/index.ts
pauljmelia: I started down a similar route, but using a lot more duplication of the base classes which I didn't like. Can't say I'm a fan of fooling the compiler but sometimes you have to wonder why methods such as `_buildServerConfig` can't be made more accessible.

joeldenning: Agreed. See related https://github.com/manfredsteyer/ngx-build-plus/pull/49 where this is what I ended up doing in ngx-build-plus to get things working.

PR #40, opened 2019-04-01
Add ability to merge external webpack config into angular-cli config

pauljmelia:
Basically I've mimicked the way that angular-builders/custom-webpack enabled an external webpack config into the angular-cli webpack config, so you can do things like add additional entry points, etc. to the single-spa app output


src/builders/webpack-config-merger.ts
```diff
@@ -0,0 +1,21 @@
+import {MergeStrategies} from './single-spa-webpack-builder-config';
+import * as webpackMerge from 'webpack-merge';
+import {Configuration} from 'webpack';
+import {differenceWith, keyBy, merge} from 'lodash'
+
+export class WebpackConfigMerger {
+    static merge(webpackConfig1: Configuration, webpackConfig2: Configuration, mergeStrategies: MergeStrategies = {}, replacePlugins = false): Configuration {
+        const mergedConfig = webpackMerge.smartStrategy(mergeStrategies)(webpackConfig1, webpackConfig2);
```
joeldenning: Why aren't we just using [webpackMerge.smart()](https://github.com/survivejs/webpack-merge#smart-merging)?  It is already premade for webpack merges and in my experience works great without any configuration. It would simplify the code in this PR a fair amount.

src/builders/browser/index.ts
```diff
@@ -17,42 +18,18 @@ export class SingleSpaBrowserBuilder extends BrowserBuilder {
...
-        libraryTarget: options.libraryTarget as any,
-      },
-      externals: {
-        'zone.js': 'Zone',
-      },
-    });
+    // Delegate the building of the webpack config to the new builder.
+    // Builder based on custom builders implmented by @meltedspark
+    // https://github.com/meltedspark/angular-builders
+    return SingleSpaWebpackBuilder.buildWebpackConfig(root, options.singleSpaWebpackConfig, config, options, this.context);
```
joeldenning: Being honest, I think that this pull request has made the code significantly more complicated than it needs to be. Here is what we could have done:

```js
const extraConfig = options.singleSpaWebpackConfig ? require(options.singleSpaWebpackConfig) : {}
return webpackMerge.smart(config, {
  output: {
    library: libraryName,
    libraryTarget: options.libraryTarget as any,
  },
  externals: {
    'zone.js': 'Zone',
  },
}, extraConfig);
```

Is there a reason why this implementation ^ wouldn't work?

src/builders/webpack-config-merger.ts
pauljmelia: My aim was to afford the single-spa app developer as much control as possible over what is to be merged and how, hence the use of smartStrategy and the fact that the singleSpaWebpackConfig node in the schema is an object and not just a string. 

I'll concede that this adds a fair amount of complexity, but as I can see that there might be instances where developers might want control of how their module rules and plugins are merged, I'd rather offer that from the get go.

src/builders/browser/index.ts
pauljmelia: I don't see why this wouldn't work if you don't want the additional flexibility. I followed the meltedspark implementation and refactored the actual merge into helper classes so that it could be re-used in other builders should we find the need to implement them.

As it stands the dev server builder is dependent on the browser build, which IMO is a bit of a smell. I should, as part of this PR, break that dependency and make the dev-server buider use the merger.

pauljmelia: Just added a commit that makes the dev server builder use the helper class rather than the browser builder

src/builders/webpack-config-merger.ts
joeldenning: Do you have a specific use case where you need the merge strategies?

I prefer starting out with something simple and only making it complex if there are strong use cases for the complexity. I've used webpack-merge a fair amount and haven't run into any issues with its smart merging.

pauljmelia:
@joeldenning Simplified PR to use smart instead of smartStrategy

src/builders/dev-server/index.ts
```diff
@@ -3,38 +3,52 @@ import { BuilderContext } from '@angular-devkit/architect';
...
+        const webpackConfig = SingleSpaWebpackBuilder.buildWebpackConfig(root, options.singleSpaWebpackConfigPath, config, options, this.context);
+        this.context.logger.info(tags.oneLine`
+        [single-spa-angular]: Angular dev server is serving application as a single module`);
         return webpackConfig;
-    }    
+    }
     _buildServerConfig(root, projectRoot, options, browserOptions) {
         // @ts-ignore
         const devServerConfig = super._buildServerConfig(root, projectRoot, options, browserOptions);
+        const contentBase = path.resolve(root, projectRoot.serveDirectory || '../');
```
joeldenning: This causes bugs in Windows. We should not use `path.resolve()` for this, since webpack needs paths that have `/` instead of `\`. See related release notes for beta 7, 8, and 9 at https://github.com/CanopyTax/single-spa-angular/releases

src/builders/dev-server/index.ts
```diff
@@ -3,38 +3,52 @@ import { BuilderContext } from '@angular-devkit/architect';
...
+        const lastPos = root.lastIndexOf('/') || root.lastIndexOf(path.sep);
```
joeldenning: Why the change here? Using path.sep has a strong chance of breaking Windows, since the root path in the angular json for Windows projects only contains `/` instead of `\`

src/builders/dev-server/index.ts
```diff
@@ -3,38 +3,52 @@ import { BuilderContext } from '@angular-devkit/architect';
...
+        const publicPath = root.slice(lastPos) + '/' + options.outputPath
+        this.context.logger.info(tags.oneLine`[single-spa-angular]: webpack output is served from ${publicPath}`);
+        this.context.logger.info(tags.oneLine`[single-spa-angular]: content not from webpack is served from ${contentBase}`);
 
         return webpackMerge.smart(devServerConfig, {
             // @ts-ignore
-            contentBase: projectRoot.serveDirectory || '../',
+            contentBase: contentBase,
```
joeldenning: Let's just keep the code as it was beforehand -- I worked really hard to get it working in all operating systems and would rather not change it without a reason to.

src/builders/dev-server/index.ts
```diff
@@ -3,38 +3,52 @@ import { BuilderContext } from '@angular-devkit/architect';
...
+        // Delegate the building of the webpack config to the new builder.
+        // Builder based on custom builders implmented by @meltedspark
+        // https://github.com/meltedspark/angular-builders
+        const webpackConfig = SingleSpaWebpackBuilder.buildWebpackConfig(root, options.singleSpaWebpackConfigPath, config, options, this.context);
+        this.context.logger.info(tags.oneLine`
+        [single-spa-angular]: Angular dev server is serving application as a single module`);
         return webpackConfig;
-    }    
+    }
     _buildServerConfig(root, projectRoot, options, browserOptions) {
```
joeldenning: We should either (1) not change this function at all (since there is no new functionality) or (2) change this function so that the extra webpack config is respected for the devServerConfig (which it currently appears not to be, since the config that we return has single-spa-angular's devServer config be the last thing merged in)

src/builders/browser/index.ts
```diff
@@ -17,42 +18,18 @@ export class SingleSpaBrowserBuilder extends BrowserBuilder {
...
-        libraryTarget: options.libraryTarget as any,
-      },
-      externals: {
-        'zone.js': 'Zone',
-      },
-    });
+    // Delegate the building of the webpack config to the new builder.
+    // Builder based on custom builders implmented by @meltedspark
+    // https://github.com/meltedspark/angular-builders
+    return SingleSpaWebpackBuilder.buildWebpackConfig(root, options.singleSpaWebpackConfigPath, config, options, this.context);
```
joeldenning: Can you add to [this part of the readme](https://github.com/CanopyTax/single-spa-angular#ng-build-options) documentation for this new option?

src/builders/dev-server/index.ts
```diff
@@ -3,38 +3,52 @@ import { BuilderContext } from '@angular-devkit/architect';
...
         // @ts-ignore
         const devServerConfig = super._buildServerConfig(root, projectRoot, options, browserOptions);
+        const contentBase = path.resolve(root, projectRoot.serveDirectory || '../');
+        const lastPos = root.lastIndexOf('/') || root.lastIndexOf(path.sep);
+        const publicPath = root.slice(lastPos) + '/' + options.outputPath
+        this.context.logger.info(tags.oneLine`[single-spa-angular]: webpack output is served from ${publicPath}`);
+        this.context.logger.info(tags.oneLine`[single-spa-angular]: content not from webpack is served from ${contentBase}`);
 
         return webpackMerge.smart(devServerConfig, {
```
joeldenning: this overwrites the custom webpack config

package.json
```diff
@@ -60,6 +60,7 @@
     "webpack": "^4.29.6"
   },
   "dependencies": {
+    "lodash": "^4.17.11",
```
joeldenning: This appears to not be used. Let's remove it.

src/builders/dev-server/index.ts
pauljmelia: Not a problem. Just as an FYI though, this has been working fine on my Windows 10 Insider Preview Skip-Ahead build.

src/builders/dev-server/index.ts
pauljmelia: Removed use of path.sep; I'd left it in as a belts-n-braces approach.

src/builders/dev-server/index.ts
pauljmelia: Done

src/builders/dev-server/index.ts
pauljmelia: My oversight. I've just hit this with my test project, while it did seem to respect the additions I'd made to the entry node, it has ignored additions I've made to the devServer node.

src/builders/dev-server/index.ts
pauljmelia: Yep... see previous comment

package.json
pauljmelia: Done

src/builders/dev-server/index.ts
pauljmelia: Phew, what a palaver! After a while of digging around, I realised that the angular-cli devServer that we're extending does correctly call into our overriding method and the merge that's done in `buildServerConfig` is only a merge of the `devServer` node. Because of the way that angular-cli is implemented, however, if you extend the `devServer.https` node in the custom config (my use case is that I need to use the `pfx` & `passphrase` properties that isn't implemented in angular cli) you still need to specify `ssl: true` in `angular.json` to get https to work.

PR #54, opened 2019-04-10
updated readme

StephenRansom:
Documentation update related to
https://github.com/CanopyTax/single-spa-angular/issues/51

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ The schematic performs the following tasks:
 
 ### Check if it works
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,6 +59,11 @@ Finally, run the following command from inside of the application directory:
 ng serve --open
 ```
 
+Alternatively you can run the following command from the root directory where your index.html is located:
```
StephenRansom: Feel free to remove this since it's not related to the component selector issue.

joeldenning: This will only work if you run `ng build` before doing `npx light-server`, so I think we should move this into the Building section.

joeldenning:
I'll make that small readme change I suggested above ^ in a separate commit.

Thanks for your help with this!

PR #73, opened 2019-05-10
Added props to bootstrapFunction

Magefo:
Hi, this change makes possible to use the customProps from the register application into the singleSpaAngular bootstrapFunction. for example: 

```javascript
bootstrapFunction: (props) => platformBrowserDynamic(
    {provide: 'environment', useValue: props.environment }
]).bootstrapModule(AppModule)
```


joeldenning:
This is released in beta 21

PR #88, opened 2019-06-24
fix browser back button issue

ernestlv:
we need to invoke ngZone.run() to kick in angular change detection cycle after browser popstate event is triggered via browser back / forward button. Otherwise Angular components are not updated.

### DIFF
--- a/src/browser-lib/single-spa-angular.ts
+++ b/src/browser-lib/single-spa-angular.ts
@@ -55,6 +55,7 @@ function bootstrap(opts, props) {
   });
 }
 
+let routingListener;
 function mount(opts, props) {
   return Promise
     .resolve()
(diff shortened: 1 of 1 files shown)


src/browser-lib/single-spa-angular.ts
```diff
@@ -55,6 +55,7 @@ function bootstrap(opts, props) {
   });
 }
 
+let routingListener;
```
joeldenning: Although uncommon, there are some people who use the same instance of the single-spa-angular library for multiple angular applications.

For those people, creating a shared variable for all of them is problematic. Instead, let's add `routingListener` to `opts`, which is the object that is specific to each angular application.

src/browser-lib/single-spa-angular.ts
```diff
@@ -78,7 +79,13 @@ function mount(opts, props) {
           throw Error(`single-spa-angular: the opts.bootstrapFunction returned a promise that did not resolve with a valid Angular module. Did you call platformBrowser().bootstrapModuleFactory() correctly?`)
         }
         module.injector.get(opts.NgZone)._inner._properties[opts.zoneIdentifier] = true;
-        
+        const ngZone = module.injector.get(opts.NgZone);
+        const ngZoneID = opts.zoneIdentifier;
+        routingListener = () => {
+          //fix browser back / forward button
+          ngZone.run(() => console.log("ngZone for", ngZoneID));
```
joeldenning: This would be quite annoying to see in the console all the time. The single-spa-angular community [is already displeased](https://github.com/CanopyTax/single-spa-angular/issues/53) about the "Navigation triggered outside of Angular" warning in the console, and so let's remove the console.log() altogether.

Issue #140, opened 2020-03-10
Parcel is not updating when custom props are changed in the host app


victorlevasseur:
Hi,

I just began to use single-spa with single-spa-angular to try the Parcel feature. It seems that the Angular parcel component is ready to allow custom props updates (see https://github.com/single-spa/single-spa-angular/blob/master/src/browser-lib/parcel-lib/parcel.component.ts#L66) but the singleSpaAngular function to create a Parcel config is not providing an update function (see https://github.com/single-spa/single-spa-angular/blob/master/src/browser-lib/single-spa-angular.ts#L37)

Should something be done to fix that? I would happily contribute to fix this as I'm currently facing this limitation.

victorlevasseur:
Here is what I added to make the update work without touching the single-spa-angular code base:
```
const LogoutButtonParcelInternal = singleSpaAngular({
  bootstrapFunction: (props) => {
    singleSpaPropsSubject.next(props);
    return platformBrowserDynamic().bootstrapModule(AppModule);
  },
  template: '<lbp-logout-button />',
  NgZone
});
LogoutButtonParcelInternal.update = (props) => { // Needed so that we receive the new props if their are updated.
  singleSpaPropsSubject.next(props);
  return Promise.resolve();
};

export const LogoutButtonParcel = LogoutButtonParcelInternal;
```

Maybe we can add an updateFunction in the singleSpaAngular function.

joeldenning:
Hi @victorlevasseur - yes let's add an `update` lifecycle to single-spa-angular. I think that calling `singleSpaPropsSubject.next()` is the easiest and best way. Thank you for being willing to a submit a PR for this - I'll be happy to review it and get it published once it's merged.

joeldenning:
Resolved in #141

PR #141, opened 2020-03-11
Add optional update lifecycle to singleSpaAngular

victorlevasseur:
Following our discussion in #140.
I just added an optional opts updateFunction that is directly set as the update lifecycle of the produced parcel/app. The user can use it to update its local singleSpaProps subject.


joeldenning:
Could you document this new feature in https://github.com/single-spa/single-spa.js.org/blob/master/website/versioned_docs/version-5.x/ecosystem-angular.md?

joeldenning:
Released in https://github.com/single-spa/single-spa-angular/releases/tag/v3.3.0

Issue #145, opened 2020-03-19
Dynamic HTMLElements as app mounting points


krzyswiatkowski:
### **What's wrong?**
`domElementGetter` expects to find DOM element immediately. Dynamically created DOM element causes throwing error: `domElementGetter did not return a valid dom element`

### **Expected behaviour**
`domElementGetter` should take into account that DOM element on which app will be mounted might be created dynamically.

### **Suggested solution**
Call `domElementGetter` with reasonable amount of retries in case of DOM element not found immediately

#### **Implemented solution**
PR: https://github.com/single-spa/single-spa-angular/pull/146

krzyswiatkowski:
@joeldenning I Would much appreciate if you could take a look on this issue & implemented solution as we're in desperate need of that fix :)

joeldenning:
You can provide your own domElementGetter that ensures the dom element is created correctly. [Here is an example](https://github.com/react-microfrontends/navbar/blob/master/src/react-mf-navbar.js#L21)

```ts
const lifecycles = singleSpaAngular({
  bootstrapFunction: singleSpaProps => {
    return platformBrowserDynamic().bootstrapModule(AppModule);
  },
  template: '<app1-root />',
  Router,
  NgZone: NgZone,
  domElementGetter
});

function domElementGetter() {
  let domElement = document.getElementById('app1');
  if (!domElement) {
    domElement = document.createElement('div')
    domElement.id = 'app1'
    document.body.appendChild(domElement)
  }
  return domElement
}
}
```

Alternative:

```ts
const lifecycles = singleSpaAngular({
  bootstrapFunction: singleSpaProps => {
    return platformBrowserDynamic().bootstrapModule(AppModule);
  },
  template: '<app1-root />',
  Router,
  NgZone: NgZone,
  domElementGetter: () => document.getElementById('app1')
});

export const bootstrap = [
  () => {
    const domEl = document.createElement('div')
    domEl.id = 'app1'
    document.body.appendChild(domEl)
  }
]
export const mount = lifecycles.mount
export const unmount = lifecycles.unmount
```

krzyswiatkowski:
@joeldenning Thank you for that, it does heelp. 
We had a case where DOMElement for our child-app was connected to parent-angular-router so when user left child-app-URL the element was destroyed (parent angular-router changed it for another one)

[...]

PR #146, opened 2020-03-19
Mounting Apps on dynamically created DOM Element

krzyswiatkowski:
### Resolves: #145 
 - SingleSpaAngularOpts accepts `domElementMaxRetry` and `domElementRetryInterval`
- `domElementMaxRetry` how many times lib will try to find DOM Element on which app should be mounted
- `domElementRetryInterval` if previous call to `domElementGetter` did not return valid HTML element; how long should wait for next try (in ms)


joeldenning:
The domElement should be created either by the domElementGetter or by a custom bootstrap lifecycle method, as shown in https://github.com/single-spa/single-spa-angular/issues/145#issuecomment-601513124. We do not need to change single-spa-angular to support this.

PR #156, opened 2020-04-06
build: setup Cypress for end-to-end testing

arturovt:
@joeldenning 

In short what I've done:
* added `integration` folder with 2 folders inside:
  * `shop` (simple Angular application with routing)
  * `portal` (`single-spa` web app)
* added cypress. Added scenario for the "back button" issue when change detection gets broken

What changes will be required further?
* currently package.json of the NPM package is located in the root folder
  * the root package.json should contain only devDependencies and be `private: true`
* NPM package.json should be isolated inside the `src` folder
* `single-spa-angular/browser-lib` should be built with ng-packagr since it emits Angular compatible code that can be proceeded by both view engine and Ivy compilers


joeldenning:
This is a big one - will try to make time for it today, but it might be tomorrow before I can look at it.

yarn.lock
```diff
@@ -0,0 +1,10776 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
joeldenning, line 1: singlee-spa-angular currently uses npm, not yarn. There is a package-lock.json file in source control and in this PR that was changed. Are you proposing changing to yarn? Or why did you modify both the package-lock and the yarn.lock?

package.json
```diff
@@ -42,28 +55,48 @@
...
     "typescript": "3.7.5",
-    "webpack": "^4.29.6"
+    "webpack": "^4.29.6",
+    "webpack-cli": "^3.3.11",
+    "zone.js": "~0.10.3"
   },
   "dependencies": {
     "single-spa": ">= 5.1.1"
+  },
+  "prettier": {
```
joeldenning, line 96: Why is this here, when single-spa-angular doesn't currently use prettier to autoformat files? I am open to adding prettier to the project, but that would necessitate installing prettier, pretty-quick, and husky as devDependencies, and then adding a pre-commit git hook. I'd rather not have a prettier config unless we're actually using prettier.

package.json
```diff
@@ -13,7 +13,20 @@
     ":copy-schematic-files": "cpx \"src/**/_files/**/**\" lib",
     "parcel": "ngc -p tsconfig.parcel.json",
     "spa": "tsc -p tsconfig.spa.json",
-    "node": "tsc -p tsconfig.node.json"
+    "node": "tsc -p tsconfig.node.json",
+    "// - Integration": "Building apps that are required for E2E testing",
+    "build:integration:shop": "yarn --cwd integration/shop build:single-spa",
```
joeldenning, line 18: Why the switch from npm to yarn? I don't really care whether this project uses npm or yarn, but feel pretty strongly that it should only use one or the other (not both). I personally prefer yarn, but the previous person helping out with single-spa-angular didn't like it so I let him change it over to npm.

integration/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+version: '3.7'
```
joeldenning, line 1: For what purpose do we need docker in the integration tests? Can't we just do `ng serve` or `npm start`?

integration/portal/webpack.config.js
```diff
@@ -0,0 +1,25 @@
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+
+module.exports = {
+  entry: './src/main.ts',
```
joeldenning, line 4: Why are we using webpack directly instead of angular cli? Wouldn't using angular cli be a more accurate representation of how people generally use single-spa-angular?

^ I say this as someone who likes using webpack directly and finds angular schematics and angular builder confusing. But if we're going for e2e tests, I would have expected angular cli since that's what most people use

cypress/integration/issue-94.js
```diff
@@ -0,0 +1,30 @@
+/// <reference types="cypress" />
+
+describe('https://github.com/single-spa/single-spa-angular/issues/94', () => {
+  it('should navigate using back button and change detection should stop working', () => {
```
joeldenning, line 4: 👍 👍 🎉

cypress.json
```diff
@@ -0,0 +1,10 @@
+{
```
joeldenning, line 1: Are you planning on adding a CI process to run these cypress tests during PRs? Or only manually? I am new to Cypress and might be missing something, but think the greatest value would be if we use Travis CI, Github Actions, or similar to run the tests for each branch / PR.

I understand that getting CI working can also be a challenge, so if you'd like to wait on that for a future PR, that's okay with me, too.

yarn.lock
arturovt: > Are you proposing changing to yarn?

Yes. I do prefer yarn over npm same as you. I'm also awaiting the yarn 2. That will significantly speed up the installation process in the future when it comes out of beta. What are your thoughts? I forgot to remove `package-lock`.

package.json
arturovt: See comment above.

integration/portal/webpack.config.js
arturovt: The `portal` is not an angular app. It's a single web-app that uses `single-spa` and gets bundled and served via webpack.

cypress.json
arturovt: > Are you planning on adding a CI process

For sure, captain! :slightly_smiling_face:

package.json
arturovt: True. That is what I'm planning to do next.

I actually have an auto-formatting in my vscode and it relies on the project config. We could leave that config here and I will add prettier/husky/etc. in the next PR. Thoughts?

yarn.lock
joeldenning: 👍 let's remove the package-lock.json and switch to yarn 1. Yarn 2 is quite a leap, with removal of the node_modules directory, so I'd prefer holding off on that one.

package.json
joeldenning: Sure, sounds good

integration/portal/webpack.config.js
joeldenning: Ah I see - I didn't realize that. Yes in that case it makes a ton of sense to do it this way

package.json
```diff
@@ -42,28 +55,48 @@
...
     "eslint": "^5.15.2",
     "jest": "^24.5.0",
     "rimraf": "^2.6.2",
+    "rxjs": "~6.5.5",
+    "start-server-and-test": "^1.10.11",
     "ts-jest": "^24.0.0",
+    "tslib": "^1.11.1",
     "typescript": "3.7.5",
-    "webpack": "^4.29.6"
+    "webpack": "^4.29.6",
```
joeldenning, line 89: Why is webpack/webpack-cli in this package.json? Shouldn't it only be in the integration/portal package.json?

integration/docker-compose.yml
arturovt: Both applications (shop and portal) are built in the production mode and webpack emit their assets to `dist` folders. We would like to have some kind of web servers that would serve those static files and listen appropriate ports. Docker is super helpful here since it requires a little bit of configuration and it's able to run both applications using just 1 command.

Might be harder to accomplish the same behaviour using `npm start` or other commands.

package.json
arturovt: Sure, will move it.

integration/docker-compose.yml
joeldenning: I would rather not require contributors and the CI to install docker just to run the tests. You can run multiple servers in a single command using [concurrently](https://www.npmjs.com/package/concurrently) or similar.

Docker is a great tool, but I think here it adds barriers to contribution, instead of removing them.

```json
{
  "scripts": {
    "start:integration": "concurrently -n w: yarn:start:integration:*",
    "start:integration:portal": "yarn --cwd ...",
    "start:integration:shop": "yarn --cwd ..."
  }
}
```

arturovt: Honestly saying I was thinking about `concurrently`. But it would require to have node.js servers in both apps to serve statics, doesn't it?

> the CI to install docker

Github actions or Circle have pre-installed docker, also Travis provides it via service (requires 1-line command).

arturovt: > I would rather not require contributors

Agreed. In my opinion, we could have both. That Docker image, that is used in the config, is a Go image with 5 MB size (super lightweight and super fast).

And also provide a way to serve statics via another approach for other people who would want to contribute.

Personally, I work mostly all-time with Docker, but it's only a subjective vision for sure :smile:

joeldenning: Gonna have to insist here - I use docker as well but think that requiring docker to run tests is a barrier to contribution. This library is already very complex (browser lib, schematics, builder, supporting different versions of angular, etc).

Running two web servers is completely fine.

arturovt: Okay, I will add concurrently then and 2 simple express servers.

joeldenning: You can use [serve](https://www.npmjs.com/package/serve) or [http-server](https://www.npmjs.com/package/http-server) to avoid writing your own static file servers. Also, you can use webpack-dev-server to host the files too (which is what I normally do).

joeldenning: Serve even has support for SPA mode, where all URLs default to returning the index.html. You can do that with `serve -s`

arturovt: Even better :)

arturovt: Removed docker and added "serve".

joeldenning:
Thanks again for all your work on this - I am signing off for the day but will look at this tomorrow morning.

Issue #94, opened 2019-07-09
Back Button Not working!!


ashuaggarwal94:
Hi,

I have been facing an issue with the single spa framework and angular routing, in which when i move from one component after performing the functionality like search and then i move to another component. And then, press the browser back button and the UI is not getting rendered correctly or functionality doesn't work the way it should.
For reproducing the issue, the below repo has been created. and do find the video in the repo.

Repo: https://github.com/ashuaggarwal94/singleSPA1

amaneer94:
I am facing the same kinda problem.

joeldenning:
I'm taking a look at this now - hopefully I can find the issue.

joeldenning:
The behavior that I'm seeing is that the back button works but that when you attempt to search again that I get an error. Is the below gif what you're seeing?

![back-button-1](https://user-images.githubusercontent.com/5524384/60907056-db6f9480-a235-11e9-986c-4af011f755c6.gif)

ashuaggarwal94:
My bad!!! Changed the code, can you please pull the code changes and try now!! 
We are not getting the error now!! even in above case we should not have got the error. But we got. As of now instead of two way binding i have declared a template variable now.

Thanks,
Ashutosh

joeldenning:
With your new changes to the code, I am no longer seeing any errors and the back button is working perfectly. I tested in both Firefox and Chrome.

![back-button-bug](https://user-images.githubusercontent.com/5524384/61475474-18c7d680-a948-11e9-8b3e-569385adebee.gif)

Is this different than what you're seeing locally?

joeldenning:
<img width="864" alt="image" src="https://user-images.githubusercontent.com/5524384/62309086-fc499500-b443-11e9-8cfe-7653664f7c12.png">

Link to repo that demos the problem: https://github.com/korjikk/angular-single-spa-routing-issue

ashuaggarwal94:
Hi @joeldenning ,

I tried downgrading the router but it had started giving me error while i was building the project.
One more thing that i have observed is we are able to run the application if we are using enter (the way u are doing) but the submit button is not working!! 

<input (keyup.enter)="onClick()" />

We had found an alternative to the problem as of now by using ChangeDetectorRef and calling detectChanges method over the call.

Eg:

constructor(private ref: ChangeDetectorRef){}

onClick(value){
this.data.push(value);
this.ref.detectChanges();
}

Thanks,
Ashutosh

rknash18:
[...]

PR #159, opened 2020-04-07
fix: provide custom `PlatformLocation` to intercept `popstate` and wrap callback into zone

arturovt:
Issue #94


src/browser-lib/life-cycles.ts
```diff
@@ -0,0 +1,124 @@
...
+
+import { SingleSpaAngularOpts, BootstrappedSingleSpaAngularOpts } from './types';
+import {
+  chooseDomElementGetter,
+  getContainerEl,
+  ivyEnabled,
+  removeApplicationFromDOMIfIvyEnabled,
+} from './dom';
+
+export async function bootstrap(opts: BootstrappedSingleSpaAngularOpts, props: any): Promise<void> {
```
joeldenning, line 11: Moving all the code to a new file makes it a lot harder for me to review this effectively - I don't know what changed and what didn't.

I'd rather stick to the pattern that every other single-spa helper library uses, and keep these lifecycles in the main file.

src/browser-lib/extra-providers.ts
```diff
@@ -0,0 +1,42 @@
...
+class SingleSpaPlatformLocation extends ɵBrowserPlatformLocation {
+  onPopState(fn: (event: LocationChangeEvent) => void): void {
+    const ngZone = inject(NgZone);
+
+    super.onPopState(event => {
+      // Wrap any event listener into zone that is specific to some application.
+      // The main issue is `back/forward` buttons of browsers, because they invoke
+      // `history.back|forward` which dispatch `popstate` event. Since `single-spa`
+      // overrides `history.replaceState` Angular's zone cannot intercept this event.
+      // Only the root zone is able to intercept all events.
```
joeldenning, line 19: 👍 good explanation. Let's also add a link to the github issue for those wanting to read more about it.

```js
// See https://github.com/single-spa/single-spa-angular/issues/94 for more detail
```

src/browser-lib/extra-providers.ts
```diff
@@ -0,0 +1,42 @@
...
+      provide: PlatformLocation,
+      useClass: SingleSpaPlatformLocation,
+      deps: [[new Inject(DOCUMENT)]],
+    },
+  ];
+}
+
+// Providers have to be exported but we don't want our users to consume
+// this class because it's meant to be private.
+export { SingleSpaPlatformLocation as ɵSingleSpaPlatformLocation };
```
joeldenning, line 42: Why the `ɵ` character? Is that a convention within the Angular community? What is the downside of just exporting it as `SingleSpaPlatformLocation`

src/browser-lib/types.ts
```diff
@@ -0,0 +1,19 @@
+import { NgZone, Type, NgModuleRef } from '@angular/core';
+import { AppProps } from 'single-spa';
+
+export interface SingleSpaAngularOpts {
```
joeldenning, line 4: Why do we need this file? The current build process outputs `.d.ts` files (https://unpkg.com/browse/single-spa-angular@3.3.0/lib/browser-lib/single-spa-angular.d.ts). Can't we just keep these types in the file that uses them?

arturovt: No problem. I will these changes to stick in the single file.

src/browser-lib/life-cycles.ts
arturovt: No problem. I will revert these changes to stick in the single file.

src/browser-lib/extra-providers.ts
arturovt: Yeah, the "theta" symbol is a convention. Same like in that class that we extend `ɵBrowserPlatformLocation`.

> What is the downside

No downside actually, it just means that class is "private" and shouldn't be consumed by our users.

src/browser-lib/single-spa-angular.ts
```diff
@@ -39,111 +41,137 @@ export default function singleSpaAngular(userOpts: SingleSpaAngularOpts): LifeCy
...
-    // See https://github.com/PlaceMe-SAS/single-spa-angular-cli/issues/33,
-    // https://github.com/single-spa/single-spa-angular/issues/47,
-    // https://github.com/angular/angular/blob/a14dc2d7a4821a19f20a9547053a5734798f541e/packages/core/src/zone/ng_zone.ts#L144,
-    // and https://github.com/angular/angular/blob/a14dc2d7a4821a19f20a9547053a5734798f541e/packages/core/src/zone/ng_zone.ts#L257
-    opts.NgZone.isInAngularZone = function() {
-      // @ts-ignore
-      return window.Zone.current._properties[opts.zoneIdentifier] === true;
-    }
+async function bootstrap(opts: BootstrappedSingleSpaAngularOpts, props: any): Promise<void> {
+  await Promise.resolve();
```
arturovt, line 49: Seems like a lot of changes, but it's only changed from `promise.resolve().then()` to `await Promise.resolve()`.

src/browser-lib/extra-providers.ts
arturovt: Removed theta symbol since the class was used in another file.

arturovt: While testing I encountered another problem.... :monkey_face: 

The `ɵBrowserPlatformLocation` was moved into another package in Angular 9, so:
```ts
import { ɵBrowserPlatformLocation } from '@angular/platform-browser` // Angular 6-8
import { ɵBrowserPlatformLocation } from '@angular/common' // Angular 9
```

They always have some breaking changes in major versions. Yeah, it sux...

What would you propose? For sure we could publish different versions `single-spa-angular@4` where peer dependency would be `@angular/core < 9` and `single-spa-angular@5` with `@angular/core > 9`.

Just thinking aloud....

joeldenning:
I'm signing off for the day, but will look at this tomorrow morning. Thanks again for your good work on this. Is this PR ready to move out of WIP into Ready For Review?

src/browser-lib/single-spa-angular.ts
joeldenning: This line is unnecessary - lifecycle functions such as bootstrap just need to return a promise, which is now being accomplished by making them async functions (which always return promises)

```suggestion
```

src/browser-lib/single-spa-angular.ts
```diff
@@ -39,111 +41,137 @@ export default function singleSpaAngular(userOpts: SingleSpaAngularOpts): LifeCy
...
+    opts.bootstrappedNgZone.run(() => {
+      // See https://github.com/single-spa/single-spa-angular/issues/86
+      // Zone is unaware of the single-spa navigation change and so Angular change detection doesn't work
+      // unless we tell Zone that something happened
     });
+  };
+}
+
+async function mount(opts: SingleSpaAngularOpts, props: any): Promise<NgModuleRef<any>> {
+  await Promise.resolve();
```
joeldenning, line 74:
```suggestion
```

src/browser-lib/single-spa-angular.ts
```diff
@@ -39,111 +41,137 @@ export default function singleSpaAngular(userOpts: SingleSpaAngularOpts): LifeCy
...
-    }
-    window.removeEventListener('single-spa:routing-event', opts.routingEventListener)
-    if (opts.AnimationEngine) {
-      /*
-      The BrowserAnimationsModule does not clean up after itself :'(. When you unmount/destroy the main module, the
-      BrowserAnimationsModule uses an AnimationRenderer thing to remove dom elements from the page. But the AnimationRenderer
-      defers the actual work to the TransitionAnimationEngine to do this, and the TransitionAnimationEngine doesn't actually
-      remove the dom node, but just calls "markElementAsRemoved()".
+async function unmount(opts: BootstrappedSingleSpaAngularOpts, props: any): Promise<void> {
+  await Promise.resolve();
```
joeldenning, line 128:
```suggestion
```

src/browser-lib/single-spa-angular.ts
```diff
@@ -39,111 +41,137 @@ export default function singleSpaAngular(userOpts: SingleSpaAngularOpts): LifeCy
...
+      `single-spa-angular: the opts.bootstrapFunction returned a promise that did not resolve with a valid Angular module. Did you call platformBrowser().bootstrapModuleFactory() correctly?`
+    );
+  }
+
+  const singleSpaPlatformLocation: SingleSpaPlatformLocation | null = module.injector.get(
+    SingleSpaPlatformLocation,
+    null
+  );
+
+  if (singleSpaPlatformLocation === null) {
```
joeldenning, line 108: Hmm do you think it would be best to require these providers to be used by throwing an error if they aren't? I guess there is no situation where someone shouldn't be using the providers?

arturovt: Good point. Since this provider is needed for routing I will add a check `&& opts.Router` (thus the library consumer also has routing in his app)

arturovt:
> Looks good, just a few comments.
> 
> One other thing - are you planning on updating the test for issue-94, now that that is merged into master? I noticed that the test seems to be asserting the wrong behavior (on purpose), so that this test can fix it?

See my comment https://github.com/single-spa/single-spa-angular/pull/159#discussion_r405848688

arturovt:
> One other thing - are you planning on updating the test for issue-94

Yes, for sure. There is another problem, let me explain that.

Currently, the library is built via `tsc`. In Angular world, it shouldn't be like that. There is a tool called [ng-packagr](https://github.com/ng-packagr/ng-packagr) which is a part of Angular CLI.

Angular 8 supports differential loading. This means that it can rely on packages that expose different ESM bundles (es5 + es2015). ES5 and ES2015 are not friendly when it comes to classes, because Angular uses ES2015 bundles during development and when some `class A {}` transpired to `function A() {}` tries to extend some `class B {}` by calling `B.call(this)` then it will throw an error.

Since Angular ecosystem continuously grows I'd want this library to rely on the standard approach of building libraries provided by Angular. What do you think?

src/browser-lib/extra-providers.ts
joeldenning: Hmmm, yeah I'd like this fix to be available for as many versions of angular as possible. I think requiring the user to pass in `ɵBrowserPlatformLocation` is the easiest way to accomplish that.

src/browser-lib/single-spa-angular.ts
joeldenning: 👍

joeldenning:
Yes, I'm open to switching to ng-packagr. I'd prefer that that be done in a separate PR, though. This bug fix is a crucial one and I am hoping to be able to port it to single-spa-angular@3 for those using Angular 2-7

arturovt:
Would need to change target in tsconfig in this PR to emit es2015, because extending will not work.

joeldenning:
Hmm well we need this bugfix for Angular 2-8. It can't just be for 9. I just created the 3.x branch at https://github.com/single-spa/single-spa-angular/commits/3.x. If we switch over the master branch (for 4.x) to use ng-packagr, we need to either do the same for 3.x or find a different way to support the class extends syntax that doesn't require ng-packagr.

arturovt:
> Hmm well we need this bugfix for Angular 2-8. It can't just be for 9. I just created the 3.x branch at https://github.com/single-spa/single-spa-angular/commits/3.x. If we switch over the master branch (for 4.x) to use ng-packagr, we need to either do the same for 3.x or find a different way to support the class extends syntax that doesn't require ng-packagr.

What if we could extend class in es5 way? Wouldn't it be a code smell?

```ts
export function getSingleSpaExtraProviders(BrowserPlatformLocation: Type<PlatformLocation>): StaticProvider[] {
  @Injectable()
  class SingleSpaPlatformLocation {
    constructor() {
      BrowserPlatformLocation.call(this);
    }
  }

  SingleSpaPlatformLocation.prototype = Object.create(BrowserPlatformLocation);
  SingleSpaPlatformLocation.prototype.constructor = SingleSpaPlatformLocation;
```

:smile: :smile:

joeldenning:
I like where your head is at 😄. Is BrowserPlatformLocation an class or is it compiled down to a function? If it can sometimes be a class in its compiled form, wouldn't it throw an error ("must call class constructor with new")?

arturovt:
> I like where your head is at . Is BrowserPlatformLocation an class or is it compiled down to a function? If it can sometimes be a class in its compiled form, wouldn't it throw an error ("must call class constructor with new")?

Good point.

What a challenge... :smile:

joeldenning:
Does Angular CLI compile node_modules? If so, could we only publish the class version of it?

arturovt:
> Does Angular CLI compile node_modules? If so, could we only publish the class version of it?

I think It requires a sanity check at least with Angular 6,7,8.

joeldenning:
Hmm, I'm not sure the best approach then.

arturovt:
> Hmm, I'm not sure the best approach then.

I could integrate ng-packagr in a separate PR. It wouldn't bring breaking changes since it's just a library packager that will emit ESM bundles in different formats. I wouldn't want to reinvent the bicycle :slightly_smiling_face:

arturovt:
@joeldenning Let's try to come to some conclusion during today and weekends because my projects also want this fix :smile:

joeldenning:
Here are some options that could work (none of them totally ideal)

1. For single-spa-angular@3, use the [Option 1]() solution instead of the custom platform
2. For single-spa-angular@3, run `tsc` twice - once with target set to es5 and once to es2015. Require users to change their import path based on what they want.
3. Convert single-spa-angular@3 to ng-packagr as well, for differential loading.

^ Any of those sound alright to you?

joeldenning:
Oops I did not mean to close this PR! Reopening now

arturovt:
> For single-spa-angular@3, use the Option 1 solution instead of the custom platform

Option 1 is patching router right?

joeldenning:
Yes, patching router location, as described in https://github.com/single-spa/single-spa-angular/issues/94#issuecomment-610390153 and https://github.com/single-spa/single-spa-angular/issues/94#issuecomment-609501664

arturovt:
Ok, let's wait your PR with formatting first, then I'll merge this branch with master.

joeldenning:
That PR is merged in. Which of the above options sounds best to you for single-spa-angular@3? Patching router location?

arturovt:
> That PR is merged in. Which of the above options sounds best to you for single-spa-angular@3? Patching router location?

Yeah. Since that code inside the Router doesn't change at all, for me it seems like the best solution right, not the best one for sure. Providing platform also is not the best solution since it's very tricky.

With great power comes great responsibility :rofl:

joeldenning:
👍 sounds good. Are you planning on switching to ng-packagr in this PR? Or what is left for this PR before review/approve? I see the build is failing - what else?

arturovt:
> Are you planning on switching to ng-packagr in this PR?

Nope, might require to do some work, gonna do that in the next PR. Don't wanna make too many changes here, will be hard to review.

> Or what is left for this PR before review/approve? I see the build is failing - what else?

Gonna patch router, fix tests and push to branch :D

arturovt:
@joeldenning UPD.

arturovt:
Seems like Cypress is happy with that change.

arturovt:
Need to add es-5 and es-2015 builds also.

joeldenning:
I think we miscommunicated - I have been thinking of the master branch as the branch for single-spa-angular@4. And the `3.x` branch for single-spa-angular@3. I expected this PR to use provide the single spa platform location, and then a different pr to the 3.x branch to patch the router.

Is that what you've been thinnking?

arturovt:
So do I understand correctly:
* in this PR we provide a custom location that is imported from `@angular/common` and this fix will be for single-spa-angular@4 and Angular 9+?
* the next PR will be with patching Router for 3.x branch and Angular <= 8, right?

joeldenning:
👍 yes exactly. Apologies for the miscommunication - I did not intend to put you through a hurdle jumping exercise with this. My priorities are just making sure this bug fix gets out for as many versions of Angular as possible.

arturovt:
Ok. let me rever these fixes to provide a custom location.

integration/shop/package.json
```diff
@@ -7,6 +7,7 @@
     "build": "ng build",
     "build:single-spa": "ng build --prod --deploy-url http://localhost:4200/",
     "serve:single-spa": "ng serve --disable-host-check --port 4200 --deploy-url http://localhost:4200/ --live-reload false",
+    "preinstall": "cp -r ../../src/browser-lib ./lib",
```
arturovt, line 10: This is temporary !!
I'm explaining: `${root}/src/browser-lib/single-spa-angular.ts` imports everything from `${root}/node_modules` but we want it to import everything from `integration/shop/node_modules`.
Will be removed in next PR with ng-packagr...

arturovt:
@joeldenning what's our next step? Project structure? Or.... ?

joeldenning:
ng-packagr, maybe?

arturovt:
> ng-packagr, maybe?

:+1:

PR #160, opened 2020-04-07
refactor: remove `semver` from dependencies

arturovt:
Joel, this check for Angular has to be removed in the future IMO, because Angular 7 LTS ends on 18 April 2020. We'd want to maintain only stable versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12383,7 +12383,8 @@
     "semver": {
       "version": "6.3.0",
       "resolved": "https://registry.npmjs.org/semver/-/semver-6.3.0.tgz",
-      "integrity": "sha512-b39TBaTSfV6yBrapU89p5fKekE2m/NwnDocOVruQFS1/veMgdzuPcnOM34M6CwxW8jH/lxEa5rBoDeUwu5HHTw=="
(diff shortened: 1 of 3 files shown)


arturovt:
> Ah I didn't know angular core exported its version like this, that is nice.
> 
> > Joel, this check for Angular has to be removed in the future IMO, because Angular 7 LTS ends on 18 April 2020. We'd want to maintain only stable versions.
> 
> I disagree, actually. One of the main pitches of single-spa is to convert an old legacy system into a single-spa application so that you can use newer frameworks for new code. We have to support older versions of Angular (at least a little bit) to help people in that situation.

That might be complicated to accomplish, I'm not saying we should drop support for older or something similar, but we have to consider that idea.

If someone wants to start a new application and as you _can use newer frameworks for new code_ then he will start that application with the latest stable version. E.g. if we would have had some old legacy then we would start with the latest React (not with 15 I'm sure :D) and not with Angular 6 or 7, that sounds pragmatic and reasonable. Fix me if I'm wrong at some moments.

joeldenning:
Yeah so I'm thinking single-spa-angular@3 will continue to exist as a way to support Angular 2-7. And single-spa-angular@4 can support Angular 8-9. As new Angular versions come out, we choose whether to release a new major version of single-spa-angular or not for them.

PR #161, opened 2020-04-08
feat: fetch latest NPM package version except of hardcoding those values

arturovt:


src/schematics/ng-add/npm.ts
```diff
@@ -0,0 +1,78 @@
...
+    req.end();
+  });
+}
+
+/**
+ * We're interested in installing the latest `single-spa-angular` version.
+ */
+export async function getSingleSpaAngularDependency(): Promise<NodeDependency> {
+  const name = 'single-spa-angular';
+  const version = await getLatestNodeVersion(name);
```
joeldenning, line 51: Couldn't we just do `require('../../../package.json).version` instead of fetching from network? Also could be nice to do it for when a specific version of schematics relies on a certain version of the single-spa-angular browser lib.

src/schematics/ng-add/npm.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * We have to install `@angular-builders/custom-webpack` version compatible with the current
+ * version of Angular. If Angular is 8 then `custom-webpack@8.4.1` has to be installed.
+ */
+export async function getAngularBuildersCustomWebpackDependency(): Promise<NodeDependency> {
+  const { VERSION } = require('@angular/core');
+  const name = '@angular-builders/custom-webpack';
+
+  const version =
+    LATEST_COMPATIBLE_CUSTOM_WEBPACK_VERSIONS[VERSION.major] || (await getLatestNodeVersion(name));
```
joeldenning, line 70: Wouldn't it make sense just to install `^8` or `^9` instead of finding the exact version? We should be able to rely on semantic versioning to provide the latest version.

```
const version = `^${VERSION.major}`
```

arturovt: Sounds business!

src/schematics/ng-add/npm.ts
arturovt: Deal!

package.json
```diff
@@ -65,5 +65,10 @@
   },
   "dependencies": {
     "single-spa": ">= 5.1.1"
+  },
+  "prettier": {
```
joeldenning, line 69: Same comment on the other PR - it doesn't make sense to me to add a prettier config unless we're actually using prettier in the project.

arturovt: True. Left a comment regarding this in another PR.

joeldenning: 👍 I'll merge this in as-is and then the PR with prettier can be added later.

arturovt: Removed it :slightly_smiling_face:

joeldenning: Oops 😄

PR #164, opened 2020-04-10
chore: format all files with prettier

joeldenning:
Format all files with prettier, so that future PRs will only have non-style-related changes.


package.json
```diff
@@ -14,6 +14,7 @@
     "parcel": "ngc -p tsconfig.parcel.json",
     "spa": "tsc -p tsconfig.spa.json",
     "build:webpack": "tsc -p tsconfig.node.json",
+    "format": "prettier './**/*' --write",
```
joeldenning, line 17: This is the command I ran to format all these files

arturovt:
https://github.com/single-spa/single-spa-angular/runs/577492821

:tada:

PR #165, opened 2020-04-10
feat: generate different scripts for different projects if they are a part of the same workspace

arturovt:


src/schematics/ng-add/add-scripts.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  return collectedScripts;
+}
+
+function getPortsFromCollectedScripts(collectedScripts: string[]): number[] {
+  const ports: number[] = [];
+
+  for (const script of collectedScripts) {
+    const match: RegExpMatchArray | null = script.match(/--port \d{4}/);
```
joeldenning, line 96: Let's support five digit port numbers, too?

```suggestion
    const match: RegExpMatchArray | null = script.match(/--port \d+/);
```

src/schematics/ng-add/add-scripts.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    // `match[0]` will be a string e.g. `--port 4200`,
+    // we split by space to get that `4200`.
+    const [, port] = match[0].split(' ');
+    ports.push(+port);
+  }
+
+  // Sorts all numbers for ascending order. For example we will get
+  // `[4200, 4201, 4202]` sorted numbers. We will need `4202` to get next port.
+  return ports.sort((a, b) => a - b);
```
joeldenning, line 110: The sort function here is unnecessary, right? Each port is already a number due to line 105, so default sort will work?

```suggestion
  return ports.sort();
```

arturovt: Honestly saying haven't used `sort` in practice, just googled for `sort ascending order` :smile:

src/schematics/ng-add/add-scripts.ts
arturovt: Deal.

src/schematics/ng-add/add-scripts.ts
arturovt: This is where I got it https://stackoverflow.com/questions/1063007/how-to-sort-an-array-of-integers-correctly

arturovt: @joeldenning upd.

joeldenning: Looks like it still has the unnecessary sort function?

joeldenning: Does it not work if you call sort() without the callback? If it works let’s remove it

arturovt: removed

arturovt:
@joeldenning ping.

arturovt:
@joeldenning what do you think if we let the user to specify a port when running `ng add`?
There will be a question right after routing and browserAnimationsModule, the default port will be 4200.

joeldenning:
👍 sure

PR #171, opened 2020-04-17
Parcel fixes

joeldenning:
This PR fixes the broken build for the parcel component. Right now parcel is compiled separately with typescript instead of ng-packagr, perhaps it would be best to switch to just including it in the main ng-packagr build?

Here are things I found when looking through and testing out the code:

- The single-spa-angular Parcel component isn't being built with ng-packagr
- When I test out single-spa-angular locally, I run into the following error:
image


package.json
```diff
@@ -95,20 +94,19 @@
     "jest": "^24.5.0",
     "lint-staged": "^10.1.3",
     "ng-packagr": "^9.1.0",
-    "npm-run-all": "^4.1.5",
```
joeldenning, line 98: I figured there was no reason to have both npm-run-all and concurrently

package.json
```diff
@@ -95,20 +94,19 @@
...
+    "start-server-and-test": "^1.11.0",
     "ts-jest": "^24.0.0",
     "tsickle": "^0.38.1",
     "tslib": "^1.11.1",
     "typescript": "3.7.5",
     "zone.js": "~0.10.3"
   },
   "dependencies": {
-    "single-spa": ">= 5.1.1"
+    "single-spa": ">= 4"
```
joeldenning, line 109: single-spa-angular@4 works with single-spa@>=4 as far as I know

tsconfig.parcel.json
```diff
@@ -3,9 +3,9 @@
   "compilerOptions": {
     "target": "es2015",
     "outDir": "./parcel",
-    "rootDir": "src/browser-lib/parcel-lib"
+    "rootDir": "src/src/parcel-lib"
   },
-  "include": ["src/browser-lib/parcel-lib/**/*"],
+  "include": ["src/src/parcel-lib/index.ts"],
```
joeldenning, line 8: Without the changes here, the parcel build just failed

PR #174, opened 2020-04-20
build: add `chat` application

arturovt:
I've only added another `chat` application. Thus we will able to switch to `destroyPlatform` and test whether multiple applications can be mounted/destroyed correctly.


joeldenning:
Actually the build failed - let's fix that first

Issue #113, opened 2019-10-31
URL Redirection Infinite Loop


AndrewGlazier:
Issue seems to relate to this comment;
https://github.com/CanopyTax/single-spa-angular/issues/94#issuecomment-520550455
Creating a new ticket as I cannot reproduce using the browser back button.

This bug is reproducible in the vanilla example for coexisting angular micro-frontends found here:
https://github.com/joeldenning/coexisting-angular-microfrontends

Reproduction steps;
Using the navigation app, trigger a route change, and then quickly trigger another nav change.

Issue is reproducible on;
IE11 (Very common)
Firefox (Mediocre-Hard)
Chrome (Rare)

Notes;
In our application we have found that having more simultaneous angular apps running in parallel triggers the issue more commonly, however this bug is reproducible with only one app actively loaded by the single-spa (however with difficultly even in IE11). (Issue does not occur at all with the app running in angular standalone without the spa)

The issue appears to be also related to Router.forRoot in some way, upon removing this from an app, the issue becomes more difficult to reproduce even with IE11, a similar effect to whether the app wasn't running.

This seems to be related to the speed of the navigation. Navigating twice very quickly causes this effect.

If struggling to reproduce, we found that having a button which changes to swap between two pages and spam clicking it causes the bug to reproduce on IE11 every time.

The navigations do not need to be between separate apps, routing between two pages on the same app causes the issue.

joeldenning:
I do not have time to work on this issue - outside contributions would be greatly appreciated.

rknash18:
@AndrewGlazier I have fixed this issue by adding the onSameUrlNavigation: 'reload' option to router module in angular.

joeldenning:
Oh wow thank you for sharing that @rknash18. It would be great if that workaround works for everyone else's issues they've been having, too. I've commented in #94 asking others to test if this works for them.

If it does, let's look into incorporating this into single-spa-angular so that all benefit from it.

ashuaggarwal94:
Hi @joeldenning , @rknash18  ,

The solution provided in the comment does not work for me I m stilling using the changeDectection ref as an alternative.

rknash18:
@ashuaggarwal94 I am not sure about the issue you are experiencing , I have fixed issue with router trying go back on forth when I clicked on a menu option.

gubertcalixto:
I identified that this loop starts (in reroute.js file)
[...]

PR #176, opened 2020-04-25
test: add end-to-end test that reproduces #113 issue

arturovt:
Just added a simple test that reproduces an issue with infinite redirects.

### DIFF
--- a/cypress/integration/issue-113.js
+++ b/cypress/integration/issue-113.js
@@ -1,5 +1,8 @@
 /// <reference types="cypress" />
 
+// Returning `false` prevents Cypress from failing the test because of `throw new Error`.
+Cypress.on('uncaught:exception', () => false);
+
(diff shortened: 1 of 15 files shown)


cypress/integration/issue-113.js
```diff
@@ -0,0 +1,33 @@
...
+      urlChangedTimes++;
+
+      if (urlChangedTimes === 15) {
+        // We actually wanted to navigate only 2 times to `/chat/rooms` and back to
+        // `/chat/groups`, but seems like it's already 15 times O_O.
+
+        // `setTimeout` is needed to do assertion after `throw`, since we can't
+        // execute any code after `throw`.
+        setTimeout(() => {
+          expect(urlChangedTimes).to.equal(15);
```
joeldenning, line 20: This will change to 1 or 2 after the fix?

PR #178, opened 2020-04-29
build(single-spa-angular): move internal functions to `single-spa-angular/internals` package

arturovt:
Internal functions are moved into a separate package. This will be possible to re-use them when implementing additional packages for `@angular/elements` and `renderComponent`.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,7 +9,7 @@
   "rules": {
     "semi": ["error", "always"],
     "indent": "off",
-    "@typescript-eslint/indent": ["error", 2],
+    "@typescript-eslint/indent": false,
     "@typescript-eslint/explicit-function-return-type": false,
     "@typescript-eslint/explicit-member-accessibility": false,
     "@typescript-eslint/no-explicit-any": false,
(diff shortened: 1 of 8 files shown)


.eslintrc
```diff
@@ -9,7 +9,7 @@
   "rules": {
     "semi": ["error", "always"],
     "indent": "off",
-    "@typescript-eslint/indent": ["error", 2],
+    "@typescript-eslint/indent": false,
```
arturovt, line 12: https://github.com/babel/babel-eslint/issues/681

arturovt:
It's not an npm package. In theory, it's called "secondary entry point". Here is very short docs: https://github.com/ng-packagr/ng-packagr/blob/master/docs/secondary-entrypoints.md

So we gonna have a single package `single-spa-angular` and sub-packages `single-spa-angular/{elements,render-component}`. Thus people will not bundle the functionality they don't need or use. If the developer wants to `bootstrap/mount/unmount` only Angular module then he will use `single-spa-angular`, if he wants to bootstrap an Angular element module that will define custom web component then he will use `single-spa-angular/elements` and etc.

Everything will be inside `single-spa-angular`. By creating sub-packages we reduce the tight coupling and add the ability to import only that stuff that you wanna use.

Issue #188, opened 2020-05-15
single-spa Not found in version 4.1.0


jimmyhdolores:
In version 3.0.1 single-spa is loaded, but now in this version 4.1.0 it no longer loads and therefore causes an error in the singleSpaPropsSubject class, note that you use the command ng add single-spa-angular

joeldenning:
`npm install --save single-spa`

single-spa-angular@3 and single-spa-angular@4 both have a peerDependency on single-spa. We made it a peer dependency because single-spa-angular@4 works with single-spa 3, 4, or 5 and we wanted to give people the option to use those versions if they are already on an older version.

The difference with single-spa-angular@4 is that single-spa is imported for its types in 4, whereas it isn't in single-spa-angular@3. There have been about ten people before you report this as a bug even though it's expected behavior. Which makes me think it was a bad choice to make it a peerDependency, even though that would make single-spa-angular@4 harder to use with newer older versions of single-spa.

Could you send in a pull request moving single-spa to be a dependency instead of a peerDependency?

MarcoBomfim:
I was just about to open the PR but noticed that single-spa is already a dependency (?): 
- [package.json:114](https://github.com/single-spa/single-spa-angular/blob/master/package.json#L114)

Or do you mean to throw it up as a devDependency?

joeldenning:
This project uses ng-packagr, which is a bit weird in that the package.json at the root of the project is not the one published to npm. See code below that shows that single-spa is a peerDependency:

https://github.com/single-spa/single-spa-angular/blob/ff2acca6fc66c12ad876d1dc883f45748822becf/src/package.json#L28

^ This is the package.json that needs to be updated.

MarcoBomfim:
I was overcomplicating things for some reason. Thanks!

joeldenning:
This was resolved in https://github.com/single-spa/single-spa-angular/pull/193. @filoxo added this as part of the angular schematics, which i think is an even better solution.

PR #190, opened 2020-05-18
chore(single-spa-angular): add single-spa as dependency 

MarcoBomfim:
## PR Type

```
[ ] Bugfix
[ ] Feature
[X] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

This PR removes single-spa as a peerDependency, and add as a regular dependency for the project.
Issue Number: #188 

## What is the new behavior?
Single-spa now is included as a dependency

```
[ ] Yes
[X] No
```


src/package.json
```diff
@@ -24,8 +24,7 @@
   "homepage": "https://github.com/single-spa/single-spa-angular#readme",
   "sideEffects": false,
   "peerDependencies": {
-    "@angular/core": ">=9",
-    "single-spa": ">=4.0.0"
```
joeldenning, line 28: This needs to be moved into the `dependencies` field within this package.json. I only see it removed from the package.json and then added to the yarn.lock. The yarn.lock addition is not sufficient, since npm/yarn install based on package.json

MarcoBomfim: Alright, I moved it to the parent package.json: https://github.com/single-spa/single-spa-angular/pull/190/files

Will update the PR today and see how it goes.

joeldenning:
The code looks fine to me, but the tests are failing. @arturovt can you comment here?

src/package.json
```diff
@@ -36,6 +35,12 @@
...
       }
     },
+    "whitelistedNonPeerDependencies": [
+      "single-spa"
+    ],
     "dest": "../lib"
+  },
+  "dependencies": {
+    "single-spa": "^5.5.1"
   }
```
arturovt, line 45: This should be removed.

arturovt: https://github.com/ng-packagr/ng-packagr/blob/master/docs/dependencies.md#why

joeldenning: But the whole point of this PR is to get rid of the peerDependency, since we've had >10 people report the peer dependency as an issue.

joeldenning: See related https://github.com/single-spa/single-spa-angular/issues/188

arturovt: Yeah, I understand the point of this PR. I'm just trying to consider all cases. The pasted link has a description of `package.dependencies` in _Why?_ section.

Wtyd?

joeldenning: My motivation is to not have to constantly answer the same question over and over again -  there have been >10 people ask me about the single-spa dependency since we released single-spa-angular@4.

I think ng-packagr's reasoning for avoiding dependencies to be weak, and ultimately a concession that Angular's ecosystem isn't very well set up for dependency management.

To me, the ideal is to somehow get users to understand that they need to install single-spa as well as single-spa-angular. We could try to do this by updating the docs to do `ng add sinlge-spa-angular single-spa`. However, I'm not confident that users will read the docs. So the alternative is to mark it as a dependency with the best possible semver:

```
"single-spa": ">=5.3.0"
```

^ This still leaves room for duplicating single-spa dependency, which is bad. But at least there aren't lots of users reporting something that's not a bug as a bug.

arturovt: Then it might be changed to `>=5.3.0`?

MarcoBomfim: Apart from updating the package.json to expect `>=5.3.0`, should any other logic be changed? I couldn't figure out why those tests fail as they do, but I'm thinking that the current tests expect it to be a peerDependency, and something should be changed to reflect single-spa as a regular dependency.

MarcoBomfim: Sure, redirects is surely more critical. If you need help testing let me know, I'll keep investigating this issue from my end.

arturovt:
I have added a check for PopStateEvents that are caused by single-spa as Joel commented in my PR. Which does `if (skipNextEvent && event.singleSpa)`.

We might consider setting `>= 5.4` then? Since `singleSpa` property was added in 5.4.

MarcoBomfim:
@joeldenning, @arturovt, should we close this one and get on with #193?

joeldenning:
Closing in favor of #193 - i think adding it via schematic is a better appraoch

PR #191, opened 2020-05-19
fix: resolve infinite redirects by ignoring non-browser PopStateEvents and cleanup resources when app gets destroyed

arturovt:
Issue: #113

### DIFF
--- a/cypress/integration/issue-113.js
+++ b/cypress/integration/issue-113.js
@@ -1,33 +1,50 @@
 /// <reference types="cypress" />
 
-// Returning `false` prevents Cypress from failing the test because of `throw new Error`.
-Cypress.on('uncaught:exception', () => false);
-
 describe('https://github.com/single-spa/single-spa-angular/issues/113', () => {
-  it('should navigate from /chat/groups to /chat/rooms and back to /chat/groups and SHOULD stuck in infinite redirects loop', () => {
+  it('should navigate from /chat/groups to /chat/rooms and back to /chat/groups and SHOULD NOT stuck in infinite redirects loop', () => {
     let urlChangedTimes = 0;
 
-    cy.on('url:changed', () => {
-      urlChangedTimes++;
-
-      if (urlChangedTimes === 15) {
-        // We actually wanted to navigate only 2 times to `/chat/rooms` and back to
-        // `/chat/groups`, but seems like it's already 15 times O_O.
-
-        // `setTimeout` is needed to do assertion after `throw`, since we can't
-        // execute any code after `throw`.
-        setTimeout(() => {
-          expect(urlChangedTimes).to.equal(15);
-        });
-
-        // We throw error by stopping those infinite redirects.
-        throw new Error('Seems like there are a lot of redirects...');
-      }
+    cy.visit('/chat/groups').then(() => {
+      // Start listening this event after navigating to `/chat/groups`.
+      cy.on('url:changed', () => {
+        urlChangedTimes++;
+      });
     });
 
-    cy.visit('/chat/groups');
     // By clicking this button we invoke the following code:
-    // `router.navigateByUrl('/rooms').then(() => router.navigateByUrl('/groups'))`
-    cy.get('button.reproduce-113-issue').click();
+    // `router.navigateByUrl('/rooms').then(() => router.navigateByUrl('/groups'))`.
+    cy.get('button.try-to-reproduce-113-issue')
+      .click()
+      .then(() => {
+        // We have navigated 2 times: to `/rooms` and back to `/groups`.
+        // Let's ensure that the queue was flushed and it's not stuck.
+        expect(urlChangedTimes).to.eq(2);
+
+        // Ensure Chrome didn't throttle navigation and we're able to navigate again.
+        cy.visit('/shop');
+        cy.get('shop-root').should('exist');
+      });
+  });
+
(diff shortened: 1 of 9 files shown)


src/src/extra-providers.ts
```diff
@@ -8,23 +8,53 @@ import {
...
+        // Only the root zone is able to intercept all events.
+        // See https://github.com/single-spa/single-spa-angular/issues/94 for more details
+        this.ngZone.run(() => fn(event));
+      }
+    };
+
+    // All listeners should be stored inside an array because the `onPopState` can be called
+    // multiple times thus we wanna reference all listeners to remove them further.
+    this.onPopStateListeners.push(onPopStateListener);
+    super.onPopState(onPopStateListener);
```
varora1406, line 57: Is it necessary to call onPopState always? what will happen if it's like below code

`
if(this.skipNextPopState) {
this.skipNextPopState = false;
} else {
const onPopStateListener = (event: LocationChangeEvent) => {
this.ngZone.run(() => fn(event());
};
this.onPopStateListeners.push(onPopStateListener);
super.onPopState(onPopStateListener);
}
`

src/src/extra-providers.ts
```diff
@@ -8,23 +8,53 @@ import {
...
-      // Wrap any event listener into zone that is specific to some application.
-      // The main issue is `back/forward` buttons of browsers, because they invoke
-      // `history.back|forward` which dispatch `popstate` event. Since `single-spa`
-      // overrides `history.replaceState` Angular's zone cannot intercept this event.
-      // Only the root zone is able to intercept all events.
-      // See https://github.com/single-spa/single-spa-angular/issues/94 for more detail
-      this.ngZone.run(() => fn(event));
-    });
+    const onPopStateListener = (event: LocationChangeEvent) => {
+      if (this.skipNextPopState) {
```
joeldenning, line 41: Does this approach cause issues when the back button is clicked? If so, try this?

```suggestion
      if (this.skipNextPopState && evt.singleSpa) {
```

More info at https://single-spa.js.org/docs/api#popstateevent

arturovt:
> When we spoke about this yesterday, you indicated that this fixed the inf redirect issue sometimes, but not always. Is that still true? Below is a suggestion that might fix the issue always.

The `destroy` method fixed it. I showed you that the callback was invoked 4 times except 2, because it was added twice. The `destroy` method has solved it.

joeldenning:
I see, that's great news!

I think it would still be best to add the `&& evt.singleSpa` check to the if statement. We only want to skip the "unnatural" PopStateEvents (the one caused by single-spa). Can you try that change out and see if it alters any behavior? My guess is that the bug will remain fixed, and that we'll avoid bugs in the future related to timing issues with popstate events.

arturovt:
> I see, that's great news!
> 
> I think it would still be best to add the `&& evt.singleSpa` check to the if statement. We only want to skip the "unnatural" PopStateEvents (the one caused by single-spa). Can you try that change out and see if it alters any behavior? My guess is that the bug will remain fixed, and that we'll avoid bugs in the future related to timing issues with popstate events.

OK. I will answer later.

src/src/extra-providers.ts
varora1406: @arturovt i can't find any change/response. Does it got missed?

arturovt: I was busy. Yes, it's a _callback_, similar to `window.addEventListener('load', onLoadListener)`.

src/src/extra-providers.ts
```diff
@@ -8,23 +8,60 @@ import {
...
-   */
   private ngZone!: NgZone;
 
+  // This is a simple marker that helps us to ignore PopStateEvents
+  // that was not dispatched by the browser.
+  private skipNextPopState = false;
+
+  private onPopStateListeners: ((event: LocationChangeEvent) => void)[] = [];
+
+  destroy(): void {
```
joeldenning, line 19: Does an instance of SingleSpaPlatformLocation ever get reused after `destroy()` is called? Or is it always reconstructed when it's remounted/rebootstrapped?

If it can get reused, we should reset `this.onPopStateListeners = []` inside of this function.

cypress/integration/issue-113.js
```diff
@@ -1,33 +1,50 @@
...
-        // We actually wanted to navigate only 2 times to `/chat/rooms` and back to
-        // `/chat/groups`, but seems like it's already 15 times O_O.
+    // By clicking this button we invoke the following code:
+    // `router.navigateByUrl('/rooms').then(() => router.navigateByUrl('/groups'))`.
+    cy.get('button.try-to-reproduce-113-issue')
+      .click()
+      .then(() => {
+        // We have navigated 2 times: to `/rooms` and back to `/groups`.
+        // Let's ensure that the queue was flushed and it's not stuck.
+        expect(urlChangedTimes).to.eq(2);
```
joeldenning, line 21: 😲 😱 

It's fixed!!

cypress/integration/issue-113.js
```diff
@@ -1,33 +1,50 @@
...
+          urlChangedTimes++;
+        });
+      });
+
+    cy.get('button.try-to-reproduce-113-issue')
+      .click()
+      .then(() => {
+        // We have navigated 2 times: to `/rooms` and back to `/groups`.
+        // Let's ensure that the queue was flushed and it's not stuck.
+        expect(urlChangedTimes).to.eq(2);
```
joeldenning, line 47: 🥴 It works!

PR #192, opened 2020-05-20
fix: provide `SingleSpaPlatformLocation` for `single-spa-angular@3.x` to resolve infinite redirects

arturovt:
Issue: #113


arturovt:
I have created a simple app with Angular 7 and copied our `integration/folder` where infinite redirects are reproduced. The issue was resolved.

src/browser-lib/single-spa-angular.ts
```diff
@@ -189,3 +210,5 @@ type SingleSpaAngularOpts = {
   domElementGetter?(): HTMLElement;
   AnimationEngine?: any;
 }
+
+export { getSingleSpaExtraProviders } from './extra-providers';
```
joeldenning, line 214: I thought that it wasn't possible to export providers like this without switching to ng-packagr?

src/browser-lib/single-spa-angular.ts
```diff
@@ -92,11 +91,33 @@ function mount(opts, props) {
...
+          // Cleanup resources, especially remove event listeners thus they will not be added
+          // twice when application gets bootstrapped the second time.
+          module.onDestroy(() => singleSpaPlatformLocation.destroy());
+        }
+
+        opts.bootstrappedNgZone = ngZone;
         opts.bootstrappedNgZone._inner._properties[opts.zoneIdentifier] = true;
         window.addEventListener('single-spa:routing-event', opts.routingEventListener)
         opts.bootstrappedModule = module;
-        patchRouter(opts);
```
joeldenning, line 99: Removing this is technically a breaking change. Before this PR, the back button bug was fixed without having to use the single-spa extra providers. But now we require users to use the extra single-spa providers.

Since it's a breaking change, it should cause a new major version of single-spa-angular@3. However, we're sort of boxed in with our versioning because single-spa-angular@4 already exists. So I think best option is to publish this as a 3.x minor and then try to document this as much as possible in release notes?

src/browser-lib/extra-providers.ts
```diff
@@ -0,0 +1,83 @@
...
+export class SingleSpaPlatformLocation extends ɵBrowserPlatformLocation {
+  private ngZone!: NgZone;
+
+  // This is a simple marker that helps us to ignore PopStateEvents
+  // that was not dispatched by the browser.
+  private skipNextPopState = false;
+
+  private onPopStateListeners: ((event: LocationChangeEvent) => void)[] = [];
+
+  destroy(): void {
```
joeldenning, line 19: We need to add `this.onPopStateListeners` to this PR, too. See related https://github.com/single-spa/single-spa-angular/pull/191/commits/f5ead6ddc461c8b4a3ac0d3214d72edff8bf2090

src/browser-lib/single-spa-angular.ts
joeldenning: Update: discussed this with Artur on slack and we sorted this out. I was mistaken.

arturovt:
Angular 6 has been verified also.

PR #193, opened 2020-05-20
feat: install single-spa dependency in angular schematic

filoxo:
## PR Checklist

- [x] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Single-spa is a peerDependency so it is not installed when the schematic is being run, which means that users have to manually install it after a project is created with create-single-spa. Its annoying at best, frustrating at worst because users don't understand why it doesn't work right after creating.

Issue Number: https://github.com/single-spa/create-single-spa/issues/106 

this issue is logged with create-single-spa but I believe it'd be better to implement here in case someone reruns the schematic or doesn't use create-single-spa. 

## What is the new behavior?

As part of the schematic, single-spa is added to the package.json dependencies.

```
[ ] Yes
[ ] No
[x] Possibly 
```

## Other information

I marked this as a breaking change because re-running this schematic may modify the version of single-spa in the target project's package.json. This may not actually be a breaking change or a big deal but I'd like to ensure that it seems ok since I don't have much experience with Angular.

### DIFF
--- a/schematics/ng-add/dependencies.ts
+++ b/schematics/ng-add/dependencies.ts
@@ -2,15 +2,20 @@ import { NodeDependencyType, NodeDependency } from '@schematics/angular/utility/
 
 interface PackageJson {
   version: string;
-  peerDependencies: string[];
-  dependencies: string[];
+  peerDependencies?: {
+    'single-spa': string;
+  };
+  dependencies?: {
+    'single-spa': string;
+  };
 }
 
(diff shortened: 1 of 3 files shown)


filoxo:
I'm now realizing that this might overlap with #190

MarcoBomfim:
@filoxo if it fixes the issue, I don't see why keep #190 open, since it is triggering some weird errors.

PR #196, opened 2020-05-22
build: target ES5 and ES2015 for `single-spa-angular@3.x` to support differential loading

arturovt:
Issue: #113

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,9 +68,11 @@
     "rimraf": "^2.6.2",
     "rollup": "^2.10.7",
     "rollup-plugin-typescript2": "^0.27.1",
-    "single-spa": "^5.5.1",
     "ts-jest": "^24.0.0",
     "typescript": "3.2.4",
     "webpack": "^4.29.6"
+  },
+  "dependencies": {
+    "single-spa": ">= 5.1.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -2,7 +2,9 @@
   "name": "single-spa-angular",
   "version": "3.4.1",
   "description": "Helpers for building single-spa applications which use Angular 2",
-  "main": "lib/browser-lib/single-spa-angular.js",
+  "main": "lib/browser-lib/cjs/single-spa-angular.js",
+  "module": "lib/browser-lib/esm/single-spa-angular.js",
+  "types": "lib/browser-lib/single-spa-angular.d.ts",
   "builders": "builders.json",
   "schematics": "./schematics.json",
   "scripts": {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -64,13 +66,11 @@
...
+    "rollup": "^2.10.7",
+    "rollup-plugin-typescript2": "^0.27.1",
+    "single-spa": "^5.5.1",
     "ts-jest": "^24.0.0",
     "typescript": "3.2.4",
     "webpack": "^4.29.6"
-  },
-  "dependencies": {
-    "semver": "^6.2.0",
-    "single-spa": ">= 5.1.1"
```
joeldenning, line 74: Why was this dependency moved to devDependencies?

package.json
```diff
@@ -2,7 +2,9 @@
   "name": "single-spa-angular",
   "version": "3.4.1",
   "description": "Helpers for building single-spa applications which use Angular 2",
-  "main": "lib/browser-lib/single-spa-angular.js",
+  "main": "lib/browser-lib/cjs/single-spa-angular.js",
+  "module": "lib/browser-lib/esm/single-spa-angular.js",
```
joeldenning, line 6: Does angular differential loading rely on the `main` and `module` fields in the package.json to differentiate between es5/es2015?

arturovt:
Checked with Angular 6.

image

package.json
arturovt: Depends on the compiler.

* View Engine relies on `fesm2015` and falls back to `module` when it doesn't exist (since we do not use ng-packagr we can just emit single ES2015 file and set `module` property)
* Ivy relies on `fesm2015_ivy_ngcc`

package.json
arturovt: I will move back to deps from devDeps.

arturovt:
Angular 8 with `ng serve` and differential loading enabled:

image

arturovt:
And Angular 8 in production mode:

image

package.json
arturovt: And the _differential loading_ is only a mechanism that does:
* reads _browserslist_ file (if exists)
* runs `require('browserslist')(...)`
* determines what browsers have to be supported
* if there are all browsers that do not support _es2015_ then target _es5_ only
* if there are all browsers that do support _es2015_ (e.g. `last 2 Chrome versions`) then target _es2015_ only
* if there are browsers that support _es2015_ & _es5_ then target _es2015_ and additionally emit _es5_ files and add scripts to HTML with `nomodule` attribute

That's it.

Issue #199, opened 2020-05-26
Error while running 'ng add single-spa-angular@3' with version 3.5


subodhkumarWM:
Steps:
Angular cli v7

```
ng new sspa-eg
ng add single-spa-angular@3
```

Gets below error
```
Path "/Users/subodhk/XPs/NgXp/S3xp/sspa/node_modules/single-spa-angular/src/schematics/ng-add/schema.json" does not exist.
Error: Path "/Users/subodhk/XPs/NgXp/S3xp/sspa/node_modules/single-spa-angular/src/schematics/ng-add/schema.json" does not exist.
    at readFile (/Users/subodhk/XPs/NgXp/S3xp/sspa/node_modules/@angular-devkit/schematics/tools/file-system-utility.js:17:15)
    at Object.readJsonFile (/Users/subodhk/XPs/NgXp/S3xp/sspa/node_modules/@angular-devkit/schematics/tools/file-system-utility.js:45:29)
    at NodeModulesEngineHost.createSchematicDescription (/Users/subodhk/XPs/NgXp/S3xp/sspa/node_modules/@angular-devkit/schematics/tools/file-system-engine-host-base.js:182:48)
    at SchematicEngine.createSchematic (/Users/subodhk/XPs/NgXp/S3xp/sspa/node_modules/@angular-devkit/schematics/src/engine/engine.js:219:38)
    at CollectionImpl.createSchematic (/Users/subodhk/XPs/NgXp/S3xp/sspa/node_modules/@angular-devkit/schematics/src/engine/engine.js:69:29)
    at AddCommand.getSchematic (/Users/subodhk/XPs/NgXp/S3xp/sspa/node_modules/@angular/cli/models/schematic-command.js:130:27)
    at AddCommand.runSchematic (/Users/subodhk/XPs/NgXp/S3xp/sspa/node_modules/@angular/cli/models/schematic-command.js:262:32)
    at AddCommand.executeSchematic (/Users/subodhk/XPs/NgXp/S3xp/sspa/node_modules/@angular/cli/commands/add-impl.js:134:31)
    at AddCommand.run (/Users/subodhk/XPs/NgXp/S3xp/sspa/node_modules/@angular/cli/commands/add-impl.js:105:21)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

joeldenning:
> Can this issue be moved to single-spa/single-spa-angular from single-spa?

Done

joeldenning:
[...]

PR #200, opened 2020-05-26
fix: do not ignore `src` when publishing because schema is located there

arturovt:
Issue: #199


.npmignore
```diff
@@ -1,6 +1,5 @@
 # Don't ignore lib
 
-src
```
joeldenning, line 3: Could we change this to the following?

```
src
!src/schematics/ng-add/schema.json
```

Or perhaps you could switch from npmignore to the package.json files, if safelisting is easier than excluding files.

joeldenning:
That change I suggested to the npmignore didn't work - I pulled down your code and ran `npm pack && tar -xzf single-spa-angular@3.5.0.tgz && ls -al package` and the src directory is still missing.

I think it would be better to switch to package.json `"files"`

joeldenning:
I updated this to use package.json files property and verified that it worked via `npm pack`

PR #207, opened 2020-06-01
feat: implement `@angular/elements` adapter

arturovt:


src/elements/src/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    const properties = await options.withProperties(options.ngModuleRef!);
+    const element = document.querySelector(options.element) as NgElement;
+
+    for (const [property, value] of Object.entries(properties)) {
+      element[property] = value;
+    }
+  }
+}
+
+function unmount(options: BootstrappedSingleSpaAngularElementsOptions) {
```
joeldenning, line 36: It's better to have the lifecycle function return a promise before throwing any errors, so that errors are a promise rejection instead of synchronous error. To do that, there are two options:

```js
async function unmount(options) {
  const node = document.querySelector(options.element)
  node.parentElement.removeChild()
}
```

or:

```js
function unmount(options) {
  return Promise.resolve().then(() => {
    const node = document.querySelector(options.element)
    node.parentElement.removeChild()
  })
}
```

src/elements/src/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  const node = document.querySelector(options.element);
+  // Removing custom element from DOM is enough since it will trigger
+  // `disconnectedCallback()` and Angular will dispose all resources.
+  node?.parentElement?.removeChild(node);
```
joeldenning, line 40: My guess is that you added the `?.` to make typescript happy. However, we want it to throw an error when `node` is null. If we're selecting the wrong element, we want the lifecycle function to reject its promise so that the single-spa application/parcel moves into SKIP_BECAUSE_BROKEN status.

Idk if typescript is smart enough to infer types from if/else statements, but if so perhaps the following could make it happy?

```js
if (node) {
  node.parentElement.removeChild()
} else {
  throw Error(`Could not find Angular element with selector ${options.element}`)
}
```

src/elements/src/types.ts
```diff
@@ -0,0 +1,23 @@
...
+  // and set input properties.
+  element: string;
+  bootstrapFunction(): Promise<NgModuleRef<any>>;
+  // It's possible to pass `@Input()` data to Angular custom element
+  // because custom element is basically a virtual component that has
+  // setters for all `@Input()` properties.
+  // Basically directly:
+  // const element = document.querySelector('app-custom-element');
+  // element.someInputProperty = 10;
+  withProperties?<P>(ngModuleRef: NgModuleRef<any>): WithProperties<P> | Promise<WithProperties<P>>;
```
joeldenning, line 16: Instead of this, I think that we should just put the single-spa custom props as the properties on the element.

src/elements/src/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  if (options.ngModuleRef !== null) {
+    return;
+  }
+
+  options.ngModuleRef = await options.bootstrapFunction();
+}
+
+async function mount(options: BootstrappedSingleSpaAngularElementsOptions) {
+  if (typeof options.withProperties === 'function') {
+    const properties = await options.withProperties(options.ngModuleRef!);
```
joeldenning, line 27: I think it would make sense to pass the single-spa props as the properties on the custom element.

src/elements/src/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  withProperties: undefined,
+  domElementGetter: undefined,
+};
+
+async function bootstrap(options: BootstrappedSingleSpaAngularElementsOptions) {
+  if (options.ngModuleRef !== null) {
+    return;
+  }
+
+  options.ngModuleRef = await options.bootstrapFunction();
```
joeldenning, line 22: How do we tell angular to mount to a specific dom element container? The domElementGetter should provide that container.

Also, why are we calling bootstrapFunction inside of single-spa's `bootstrap` lifecycle instead of `mount` lifecycle? Doesn't calling `bootstrapFunction` mount the angular application to a dom element? If, so we definitely need to move that to the `mount` function.

arturovt: AFAIK the `bootstrap` lifecycle is called once and `mount` is called every time the app is rendered. Modules that expose custom elements should be bootstrapped once.

src/elements/src/types.ts
arturovt: That has a more broad use case. The thing is that this module will be bootstrapped only once. The user may want to receive some data from the server before rendering custom web component and then set that data as `@Input()` parameters.

src/elements/src/index.ts
joeldenning: 👍 sounds good

arturovt:
The list of users is a custom element.

result

arturovt:
> This looks good to me. Thanks for your work on this. Feel free to merge. What is your username on npmjs.com? I will give you publish access so you can publish a new minor for this.

overthesanity

joeldenning:
I have given you publish permissions.

arturovt:
Will make a PR to single-spa docs website repo with the description of this functionality.

Issue #220, opened 2020-06-11
Webpack config has hard-coded "app3" library name


KevinCathcart:
The Webpack configuration has a hard-coded library name of "app3" as seen here:

https://github.com/single-spa/single-spa-angular/blob/6e3f36c22afbf5a75526aa5bf8e140dff6f85553/src/webpack/index.ts#L5-L9

That is both really weird, and is likely to cause problems in some scenarios.

It definitely causes problems if one decides they want a `libraryTarget` of `system` instead of `umd`, (perhaps because one would rather not pull in the amd extra for SystemJs), because this will force a named registration, requiring the named extra.

Not to mention the whole registering different modules with the same name thing under umd's amd mode or when using named `System.register` syntax.

Last but not least, this sets the webpack JSONP function to be `webpackJsonpapp3`, for all apps.

Looks like [this was noticed](https://github.com/single-spa/single-spa-angular/commit/52f55ce62e831274d0df05639f15b7b8e0580f27#r36137959) a while back, but did not get resolved. 

As reported there putting "libraryName" directly under Options fails due to builder schema validation failure. Custom properties passed to custom webpack builder must be placed in the "customWebpackConfig" object, which allows for unknown properties.

Of course, even better would be some way to actually get the Angular project name, and use that. There is a way, but only when using `@angular-builders/custom-webpack` version 9. If using that, the custom webpack config function gets a third argument, which includes the target project name, which could be passed into `singleSpaAngularWebpack()`. Sadly, this only works for Angular 9.

joeldenning:
Agreed that's weird - I'm not sure when/how that snuck in. Let's change it. I just verified that this is in both single-spa-angular@3 and single-spa-angular@4.

> Of course, even better would be some way to actually get the Angular project name, and use that. There is a way, but only when using @angular-builders/custom-webpack version 9. If using that, the custom webpack config function gets a third argument, which includes the target project name, which could be passed into singleSpaAngularWebpack(). Sadly, this only works for Angular 9.

Can we not get the Angular project name via something like this?

```ts
require(path.join(process.cwd(), 'angular.json'))
```

Would that perhaps not work with Angular Workspaces?

joeldenning:
> Not to mention the whole registering different modules with the same name thing under umd's amd mode or when using named System.register syntax.

Yes for systemjs usage, the output.library  option should be deleted via custom webpack builder

KevinCathcart:
[...]

PR #229, opened 2020-06-21
Library Name

KevinCathcart:
Sorry for the delay in getting this put together.

This pull request incorporates two different approaches. The two different commits show each of the approaches.

The first approach is simply a string parameter passed to `singleSpaAngularWebpack`. To be backwards compatible, it falls back to trying to get the project name from angular.json. If that fails, it uses a default value, and prints a warning message that explains why a default value might be undesirable, and a link to information on how to fix it.

The second approach is similar, but uses a property defined in angular.json. It has equivalent fallback to the first approach. For this second approach, I've also shown an example of another property (`libraryTarget`) that might be helpful, especially since that was supported back pre-Angular 8 builder.

Once we finalize the approach we want to take, and get this into mergeable shape for the 4.x branch series, I'll backport it to 3.x. (See the 3.xLibraryName branch in my fork if desired).

## PR Checklist

Not clear what the best way to even approach automated testing would be. Trying to parse the webpack output seems like a bad idea. I suppose technically setting something up to try to pull it in via named amd could work, but that feels obnoxious at best.

## PR Type

- [x] Feature

Issue Number: #220 

- [x] No

### DIFF
--- a/schematics/ng-add/index.ts
+++ b/schematics/ng-add/index.ts
@@ -33,6 +33,8 @@ import {
 interface CustomWebpackBuilderOptions extends BrowserBuilderOptions {
   customWebpackConfig: {
     path: string;
+    libraryName?: string;
+    libraryTarget?: string;
   };
 }
 
@@ -95,7 +97,7 @@ export function updateConfiguration(options: NgAddOptions) {
     const workspacePath = getWorkspacePath(host);
 
     if (atLeastAngular8()) {
-      updateProjectNewAngular(context, clientProject);
+      updateProjectNewAngular(context, clientProject, project.name);
       updateTSConfig(host, clientProject);
     } else {
       updateProjectOldAngular(context, clientProject, project);
@@ -130,7 +132,11 @@ function updateProjectOldAngular(
   clientProject.architect!['single-spa-serve'].options.browserTarget = `${project.name}:single-spa`;
 }
 
-function updateProjectNewAngular(context: SchematicContext, clientProject: WorkspaceProject): void {
+function updateProjectNewAngular(
+  context: SchematicContext,
+  clientProject: WorkspaceProject,
+  projectName: string,
+): void {
   context.logger.info('Using @angular-devkit/custom-webpack builder.');
 
   const buildTarget = clientProject.architect!.build!;
@@ -140,6 +146,8 @@ function updateProjectNewAngular(context: SchematicContext, clientProject: Works
   buildTarget.options.main = join(clientProject.root, 'src/main.single-spa.ts');
   (buildTarget.options as CustomWebpackBuilderOptions).customWebpackConfig = {
     path: join(clientProject.root, 'extra-webpack.config.js'),
+    libraryName: projectName,
+    libraryTarget: 'umd',
   };
 
   updateConfigurationsAndDisableOutputHashing(clientProject);
--- a/src/webpack/index.ts
+++ b/src/webpack/index.ts
@@ -1,11 +1,16 @@
 import * as webpackMerge from 'webpack-merge';
 import * as path from 'path';
+import * as fs from 'fs';
+import { findUp } from '@angular/cli/utilities/find-up';
 
-export default (config: any) => {
+export default (config: any, options?: any) => {
+  const libraryName = getLibraryName(options);
+
   const singleSpaConfig: any = {
     output: {
-      library: 'app3',
-      libraryTarget: 'umd',
+      library: libraryName,
+      libraryTarget: options?.customWebpackConfig?.libraryTarget ?? 'umd',
+      jsonpFunction: 'webpackJsonp' + libraryName,
     },
     externals: ['zone.js'],
     devServer: {
@@ -27,6 +32,11 @@ export default (config: any) => {
   };
 
   const mergedConfig: any = webpackMerge.smart(config, singleSpaConfig);
+
+  if (mergedConfig.output.libraryTarget === 'system') {
+    // Don't used named exports when exporting in System.register format.
+    delete mergedConfig.output.library;
+  }
 
   removePluginByName(mergedConfig.plugins, 'IndexHtmlWebpackPlugin');
   removeMiniCssExtract(mergedConfig);
@@ -74,4 +84,43 @@ function removeMiniCssExtract(config: any) {
       }
     }
   });
+}
+
+function getLibraryName(options: any) {
+  if (options?.customWebpackConfig?.libraryName) {
+    return options.customWebpackConfig.libraryName;
+  }
+
+  const projectName = getProjectNameFromAngularJson();
+  if (projectName) return projectName;
+
+  console.warn(
+    'Warning: single-spa-angular could not determine a library name to use and has used a default value.',
+  );
+  console.info('This may cause issues if this app uses code-splitting or lazy loading.');
+  if (!options) {
+    console.info('You may also need to update extra-webpack.config.json.');
+  }
+  console.info(
+    'See <https://single-spa.js.org/docs/ecosystem-angular/#use-custom-webpack> for information on how to resolve this.',
+  );
+
+  return 'angular_single_spa_project';
+}
+
+function getProjectNameFromAngularJson(): string | null {
+  const angularJsonPath = findUp(['angular.json', '.angular.json'], process.cwd());
+  if (!angularJsonPath) return null;
+
+  const angularJson = JSON.parse(fs.readFileSync(angularJsonPath, 'utf8'));
+  if (!angularJson.projects) return null;
+
+  var projects = Object.keys(angularJson.projects);
+
+  // if there is exactly one project in the workspace, then that must be this one.
+  if (projects.length === 1) return projects[0];
+
+  // If we reach here there are multiple (or zero) projects in angular.json
+  // we cannot tell which one to use, so we will end up using the default.
+  return null;
 }
--- a/src/webpack/index.ts
+++ b/src/webpack/index.ts
@@ -115,7 +115,7 @@ function getProjectNameFromAngularJson(): string | null {
   const angularJson = JSON.parse(fs.readFileSync(angularJsonPath, 'utf8'));
   if (!angularJson.projects) return null;
 
-  var projects = Object.keys(angularJson.projects);
+  const projects = Object.keys(angularJson.projects);
 
   // if there is exactly one project in the workspace, then that must be this one.
   if (projects.length === 1) return projects[0];
--- a/schematics/ng-add/_files/extra-webpack.config.js.template
+++ b/schematics/ng-add/_files/extra-webpack.config.js.template
@@ -1,7 +1,7 @@
 const singleSpaAngularWebpack = require('single-spa-angular/lib/webpack').default;
 
-module.exports = config => {
-  const singleSpaWebpackConfig = singleSpaAngularWebpack(config);
+module.exports = (config, options) => {
+  const singleSpaWebpackConfig = singleSpaAngularWebpack(config, options);
 
   // Feel free to modify this webpack config however you'd like to
   return singleSpaWebpackConfig;
--- a/schematics/ng-add/tests/index.spec.ts
+++ b/schematics/ng-add/tests/index.spec.ts
@@ -129,7 +129,11 @@ describe('ng-add', () => {
     const expectedCustomWebpackConfigPath = normalize(
       'projects/ss-angular-cli-app/extra-webpack.config.js',
     );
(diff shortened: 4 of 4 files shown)


src/webpack/index.ts
```diff
@@ -75,3 +84,55 @@ function removeMiniCssExtract(config: any) {
...
+  console.warn(
+    'Warning: single-spa-angular could not determine a library name to use and has used a default value.',
+  );
+  console.info('This may cause issues if this app uses code-splitting or lazy loading.');
+  if (!options) {
+    console.info('You may also need to update extra-webpack.config.json.');
+  }
+  console.info('See <SomeUrl> for information on how to resolve this.');
+
+  return 'angular_single_spa_project';
```
KevinCathcart, line 105: I wanted something better than 'app3' as the default. I'm certainly willing to accept suggestions for a better default.

src/webpack/index.ts
```diff
@@ -75,3 +84,55 @@ function removeMiniCssExtract(config: any) {
...
+    return options.customWebpackConfig.libraryName;
+  }
+
+  const projectName = getProjectNameFromAngularJson();
+  if (projectName) return projectName;
+
+  console.warn(
+    'Warning: single-spa-angular could not determine a library name to use and has used a default value.',
+  );
+  console.info('This may cause issues if this app uses code-splitting or lazy loading.');
```
KevinCathcart, line 99: Using .info() for the follow up lines instead of .warn(), because multiple lines of bold yellow only makes things hard to read.

src/webpack/index.ts
```diff
@@ -75,3 +84,55 @@ function removeMiniCssExtract(config: any) {
...
+  if (!options) {
+    console.info('You may also need to update extra-webpack.config.json.');
+  }
+  console.info('See <SomeUrl> for information on how to resolve this.');
```
KevinCathcart, line 103: I think that linking to an explanation would work better than trying to include it in the console output.
Of course, this explanation still needs to be written, and the url determined.

schematics/ng-add/index.ts
```diff
@@ -140,6 +146,8 @@ function updateProjectNewAngular(context: SchematicContext, clientProject: Works
   buildTarget.options.main = join(clientProject.root, 'src/main.single-spa.ts');
   (buildTarget.options as CustomWebpackBuilderOptions).customWebpackConfig = {
     path: join(clientProject.root, 'extra-webpack.config.js'),
+    libraryName: projectName,
+    libraryTarget: 'umd',
```
KevinCathcart, line 150: By changing this line, it is possible to change the default libraryTarget (perhaps to system), without it being a breaking change, as previously created apps would still use fallback default in webpack/index.ts.

joeldenning:
@arturovt do you have time to review this one?

src/webpack/index.ts
joeldenning: Perhaps we could add some documentation to the Angular page at https://github.com/single-spa/single-spa.js.org/blob/master/website/versioned_docs/version-5.x/ecosystem-angular.md?

src/webpack/index.ts
```diff
@@ -75,3 +84,55 @@ function removeMiniCssExtract(config: any) {
...
+
+function getProjectNameFromAngularJson(): string | null {
+  const angularJsonPath = findUp(['angular.json', '.angular.json']);
+  if (!angularJsonPath) return null;
+
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const angularJson = require(angularJsonPath);
+  if (!angularJson.projects) return null;
+
+  var projects = Object.getOwnPropertyNames(angularJson.projects);
```
joeldenning, line 116: Wouldn't `Object.keys()` work fine here? JSON files don't have non-enumerable properties, nor do they have prototype inheritance, so I don't think we'd lose anything by using the more common and familiar Object.keys function.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys

src/webpack/index.ts
```diff
@@ -75,3 +84,55 @@ function removeMiniCssExtract(config: any) {
...
+  // If we reach here there are multiple (or zero) projects in angular.json
+  // we cannot tell which one to use, so we will end up using the default.
+  return null;
+}
+
+function findUp(files: string[]): string | null {
+  let current = process.cwd();
+  const root = path.parse(current).root;
+
+  while (current !== root) {
```
joeldenning, line 129: This looks like an infinite loop - we are never updating `current` inside of the loop.

src/webpack/index.ts
```diff
@@ -75,3 +84,55 @@ function removeMiniCssExtract(config: any) {
...
+  var projects = Object.getOwnPropertyNames(angularJson.projects);
+
+  if (projects.length === 1) return projects[0];
+
+  // If we reach here there are multiple (or zero) projects in angular.json
+  // we cannot tell which one to use, so we will end up using the default.
+  return null;
+}
+
+function findUp(files: string[]): string | null {
```
arturovt, line 125: From my subjective perspective, this seems like a complicated implementation.

We have to be using tools that Angular DevKit already provides and you can work with Angular config using native tools:
```js
const { workspaces } = require('@angular-devkit/core');
const { NodeJsSyncHost } = require('@angular-devkit/core/node');

const host = workspaces.createWorkspaceHost(new NodeJsSyncHost());

workspaces.readWorkspace(`${process.cwd()}/angular.json`, host).then(({ workspace }) => {
  console.log(workspace.projects.keys());
});
```

It's also worth adding comments to the code. I've read it multiple times and I can't understand what's going on in `while + nested for` :smile:
Excuse my pedantry :)

src/webpack/index.ts
KevinCathcart: Agreed. That seemed like the most reasonable place to put it.

src/webpack/index.ts
KevinCathcart: Yeah `Object.keys` should be fine. I normally reach for `getOwnPropertyNames` out of habit since it is usually the safest option unless I need prototypical inheritance.

src/webpack/index.ts
KevinCathcart: Yeah, the while loop is broken. I must have accidentally lost a line.

src/webpack/index.ts
KevinCathcart: Yeah, the while loop is broken. I must have accidentally lost a line.

The problem I am trying to tackle here is that there is zero guarantee that angular.json is in the same folder as `process.cwd()`. This can happen if you run `ng build` from inside a source subfolder.
In order to find the workspace folder, upward searching is required.

All that is required before you can even call `.readWorkspace()`. 

angular-cli does its equivalent findUp logic here: https://github.com/angular/angular-cli/blob/0070ea46e6eaa25ad49d17c24fe7bed546dc1ec0/packages/angular/cli/utilities/project.ts#L29.

Once we have located the file, then yeah, we *could* use the devkit workspaces, except this is 100% synchronous code, and dev-kit workspaces is async.

src/webpack/index.ts
KevinCathcart: Fixed in latest push.

src/webpack/index.ts
arturovt: Webpack config can export a Promise or a factory that returns Promise thus you're able to work with sync/async code.

KevinCathcart:
Fixed several of the mentioned things.

Also switched the `require(angularJsonPath)` to `JSON.parse(fs.readFileSync(angularJsonPath, 'utf8'))` since that is more explicit about what we are doing, and avoids a lint suppression.

src/webpack/index.ts
KevinCathcart: I was thinking a new section before "Use Angular Builder" under manual installation could briefly document manually setting up the custom-webpack builder and document the new options, and we could point the users to that section.

src/webpack/index.ts
KevinCathcart: Yeah, but that defeats the whole point of the angular.json parsing, which is as a fallback that just works for the majority of all users if they are upgrading from an older version of the library. 

If the users need to update the `extra-webpack.config.ts` to make it export an promise returning function, then they could instead update it to pass the options argument to `singleSpaAngularWebpack` again, and update angular.json to specify the new 'libraryName' option. The new option (automatically populated by the schematic on new projects) is the primary source of the library name, with the angular.json parsing fallback as secondary, and a static value + warning as the third.

We could skip all angular.json parsing, but if we did so, then all existing users would either start getting the warning about using the default, and would need to make changes to silence it. The current approach is that only those few users who have multi-project workspaces will get the warning in the first place, as for everybody else we can infer it from the one-and-only project in angular.json.

arturovt: Okay. Can't we re-use the existing function then and not copy it from their source code?:
```js
const { findUp } = require('@angular/cli/utilities/find-up');
```

KevinCathcart: It was a from scratch re-write, not a copy.

But yeah, we could do that, with the caveat that it is not documented as a public api, so could potentially break in the future. (On the other hand custom builders were not documented as a public API before angular 8, and that was not a considered a problem for this package for angular 7.)

I'm perfectly happy to switch to that if that's what you want to do.

src/webpack/index.ts
joeldenning: I think the Use Angular Builder section of the docs refers to the Angular 7 builder, not to the custom webpack builder. I think that this could be near the section about what the schematics do?

src/webpack/index.ts
joeldenning: I vote we switch over to `@angular/cli/utilities/find-up`

joeldenning:
Any update on this? Switching to `@angular/cli/utilities/find-up` seems to be the last item to be addressed?

KevinCathcart:
Yeah. Sorry, I've gotten a bit busy, and have not had a chance to make that
change and push it. I'll try to make time for it this evening.

KevinCathcart:
I've pushed up a simplified version that uses '@angular/cli/utilities/find-up'.

The only reason I've still left this marked as a draft is that it links to a part of the documentation that does not yet exist:
https://single-spa.js.org/docs/ecosystem-angular/#use-custom-webpack

This would be a new section similar to the existing "Use Angular Builder" that documents these two options from angular.json, and explains how to make sure you are passing the options through in `custom-webpack.config.js`, probably by simply showing these two lines from the template:
```
module.exports = (config, options) => {
  const singleSpaWebpackConfig = singleSpaAngularWebpack(config, options);
```
Or even quoting the whole template since it is only 8 lines long.

I won't have a chance to draft up that documentation section tonight.

arturovt:
Hey @KevinCathcart 
Can I ask you not to squash commits? It's really hard to review what has been changed since the last commit and also it doesn't bring any benefit since we squash them when merging.

joeldenning:
@arturovt do you have any other comments on this before we merge?

KevinCathcart:
Ok cool. I'd be happy to create a PR that cherry picks this into 3.x this
evening. Having already tried that, I know it is not quite a clean merge,
but there is no significant difficulty there. (Biggest issue is some of the
files moved in a way that git failed to detect. Also a a few small linting
and lack of type annotation differences.)

joeldenning:
I put together the documentation for this at https://github.com/single-spa/single-spa.js.org/pull/293. @KevinCathcart could you review that pull request?

src/webpack/index.ts
```diff
@@ -75,3 +85,42 @@ function removeMiniCssExtract(config: any) {
...
+}
+
+function getProjectNameFromAngularJson(): string | null {
+  const angularJsonPath = findUp(['angular.json', '.angular.json'], process.cwd());
+  if (!angularJsonPath) return null;
+
+  const angularJson = JSON.parse(fs.readFileSync(angularJsonPath, 'utf8'));
+  if (!angularJson.projects) return null;
+
+  var projects = Object.keys(angularJson.projects);
```
arturovt, line 118:
```suggestion
  const projects = Object.keys(angularJson.projects);
```

joeldenning:
👍 I'll merge and publish this tomorrow morning.

joeldenning:
This is released in https://github.com/single-spa/single-spa-angular/releases/tag/v4.3.1

@KevinCathcart if you're up to PR'ing the `3.x` branch for Angular versions 2-8, that would be greatly appreciated! Thanks again for your work on this - was a high quality PR.

joeldenning:
Oops I just saw https://github.com/single-spa/single-spa-angular/pull/236 where you have already created the PR for the 3.x branch. Thanks!

KevinCathcart:
The 3.x version is PR #236

PR #235, opened 2020-07-02
fix: skip location change for non-imperative navigations

arturovt:
I've added another _navbar_ application.

The CI should fail right now since _113 issue_ tests are no passing anymore.


joeldenning:
Tests failed

arturovt:
> Tests failed

See description ;)

arturovt:
@joeldenning 

I've tried that change that you proposed:

> The skipNextPopState variable should be changed to an integer instead of a boolean. It should be incremented when pushState/replaceState are called, and decremented whenever onPopState is called. We should only call ngZone.run() when it is 0

I've changed the code to:

image

Is this correct or not?

arturovt:
Or this one?

image

joeldenning:
skipNextPopState should be decremented/incremented every time. Your first screenshot looks correct.

arturovt:
I've tried both and both of them broke all tests completely, e.g. that _back button_ issue that we had.

joeldenning:
Interesting - perhaps the fix isn't that good, then. This was based on a solution someone on slack proposed - I'll try to find that convo and get them involved in the Github issue.

arturovt:
@joeldenning 

After my night debugging I've noticed some moments.

First of all the behavior is obvious because if we've got 2 simultaneously existing apps:
* navbar
* chat (where we have that infinite loop test)

They both listen to `popstate` event and call `replaceState`. Calling `replaceState` triggers `popstate` event, so both of our applications stuck in infinite loop.

So my pop-up question is why use routing in a navbar application? Just thinking aloud. The navbar application shouldn't initialize router. I have removed `RouterModule` and `Router` from navbar app and used this:

image

I might consider to use Angular routing for navigation inside an application, but not for cross-app navigation, because unfortunately this is how Angular works :man_shrugging: 

So IMHO this is not an issue for me since solutions exist to solve that kind of problem. We could add that point to single-spa docs.

joeldenning:
> They both listen to popstate event and call replaceState. Calling replaceState triggers popstate event, so both of our applications stuck in infinite loop.

Why do they call `history.replaceState()` inside of their popstate listeners? That is unusual.

> So my pop-up question is why use routing in a navbar application?

Navbar applications generally use routing because they use `<router-link>`. It's nice to use the framework-specific router link component for cross microfrontend navigation.

arturovt:
> Why do they call history.replaceState() inside of their popstate listeners? That is unusual.

Step by step:

1) `setUpLocationChangeListener`  https://github.com/angular/angular/blob/b769771d60cdc6df18e750439bb3c10da278a4e7/packages/router/src/router.ts#L911
2) `scheduleNavigation` https://github.com/angular/angular/blob/b769771d60cdc6df18e750439bb3c10da278a4e7/packages/router/src/router.ts#L1235
3) `setBrowserUrl`
https://github.com/angular/angular/blob/b769771d60cdc6df18e750439bb3c10da278a4e7/packages/router/src/router.ts#L575
4) `history.replaceState`
https://github.com/angular/angular/blob/b769771d60cdc6df18e750439bb3c10da278a4e7/packages/router/src/router.ts#L1308

1. the user clicks on some link in navbar app
2. the link click listener invokes `router.navigateByUrl(...)`
3. `navigateByUrl` invokes `scheduleNavigation`
4. `scheduleNavigation` invokes `setBrowserUrl` which invokes `history.replaceState`
5. calling `replaceState` it dispatches `popstate` event
6. chat app listens to `popstate` event and calls `scheduleNavigation` which calls `setBrowserUrl -> history.replaceState`
7. navbar again captures `popstate` event
8. loop

joeldenning:
Thanks for walking me through this - I think a potential solution just occurred to me, although there are a few things to explore with it. The idea would be to either override `setBrowserUrl` from router.ts, or patch `history.replaceState`. In the override, you only forward it onto the real browser implementation if the URL, state, or page title have changed.

I see [here](https://github.com/angular/angular/blob/b769771d60cdc6df18e750439bb3c10da278a4e7/packages/router/src/router.ts#L1308) that setBrowserUrl always creates a new state object, but that perhaps the navigationId changes every time. The navigationId might be a problem since it's the same name (`"navigationId"`) that is reused for multiple angular applications. But if we can figure that problem out, it might help? What do you think - would that approach solve it?

doxingg:
there has a solution can resolve this infinite redirects , @joeldenning @arturovt try this

in the angular application which has router
in the root component `app.component.ts`
example in the `chat` application

```ts
import { Component, ChangeDetectionStrategy, OnInit } from '@angular/core';
import { NavigationStart, Router } from '@angular/router';
import { filter } from 'rxjs/operators';

@Component({
  selector: 'chat-root',
  templateUrl: './app.component.html',
  changeDetection: ChangeDetectionStrategy.OnPush,
})
export class AppComponent implements OnInit {
  constructor(private router: Router) {}

  ngOnInit(): void {
     this.skipLocationChangeSingleSpaAngular(this.router);
  }

  goToChatGroups(): void {
    this.router.navigateByUrl('/groups');
  }

 /**
  * reproduce infinite redirects with simultaneously existing apps
  */
  private skipLocationChangeSingleSpaAngular(router: Router) {
    // intercept NavigationStart set extras.skipLocationChange=false and currentNav.extras.replaceUrl =false
    // when currentNavigation.trigger !== 'imperative'
    // so, when navigation end the router do not call setBrowserUrl then the loop has break
    router.events.pipe(filter(s => s instanceof NavigationStart)).subscribe(s => {
      const currentNav = router.getCurrentNavigation();
      if (currentNav.trigger !== 'imperative') {
        currentNav.extras.skipLocationChange = true;
        currentNav.extras.replaceUrl = false;
      }
    });
  }
```

all angular application which has router need call this method `skipLocationChangeSingleSpaAngular` in the `app.component.ts`

I'm not sure if this could cause other problems, but it works.
hope this helpfull

joeldenning:
Thanks for sharing @doxingg - I am busy for the next day or so but will try to look in the next few days. If you have time, @arturovt, that would also be appreciated.

joeldenning:
Hhi @doxingg - I looked over your code changes and they look promising. I think we could incorporate it into single-spa-angular so that the change doesn't have to be made in everyone's code. We are already passing in `Router` as an option to `singleSpaAngular({Router, ...})`, so we likely could do it by getting the router instance from the module injector and then doing the `router.events.pipe()` inside of there.

It looks like your change blocks all non-imperative route changes. Can you confirm whether things work when clicking on a RouterLink, the back button, and the forward button?

src/src/single-spa-angular.ts
```diff
@@ -42,6 +43,10 @@ export function singleSpaAngular<T>(userOptions: SingleSpaAngularOptions<T>): Li
     throw Error(`single-spa-angular must be passed the NgZone option`);
   }
 
+  if (options.Router && !options.NavigationStart) {
```
joeldenning, line 46: This makes this change a breaking change - was that your intent? If we don't plan on this being a semver major, we should not throw this error, but rather only use NavigationStart when it's provided.

src/src/single-spa-angular.ts
```diff
@@ -187,3 +197,20 @@ async function unmount(options: BootstrappedSingleSpaAngularOptions, props: any)
   // https://github.com/angular/angular/issues/36449
   removeApplicationFromDOMIfIvyEnabled(options, props);
 }
+
+function skipLocationChangeOnNonImperativeRoutingTriggers(
```
joeldenning, line 201: Let's add a code comment to the relevant issue(s) for the bug this fixes.

joeldenning (approved):

doxingg:
> Hi @doxingg - I looked over your code changes and they look promising. I think we could incorporate it into single-spa-angular so that the change doesn't have to be made in everyone's code. We are already passing in `Router` as an option to `singleSpaAngular({Router, ...})`, so we likely could do it by getting the router instance from the module injector and then doing the `router.events.pipe()` inside of there.
> 
> It looks like your change blocks all non-imperative route changes. Can you confirm whether things work when clicking on a RouterLink, the back button, and the forward button?

I think it's a good idea to pass router in option. I test with chrome@84.0.4147.105、edge@86.0.622.51、firefox@82.0.2,  no problem, it work fine.

joeldenning:
Great - thanks @doxingg for your help with this! We will publish a new version of single-spa-angular shortly

joeldenning:
This is released in https://github.com/single-spa/single-spa-angular/releases/tag/v4.6.0

PR #236, opened 2020-07-03
Library name [3.x]

KevinCathcart:
3.x branch version of library name fix.

I had to add '@angular/cli' as a dev dependency, as otherwise typescript would break trying to type-check the imported findUp method. The file that needs it is only used by angular cli version 8 or newer, so we can be sure the user will have that package installed at runtime. I did however use version 7 in the dev dependency, so as to be consistent with the angular version used with the rest of the package. This is fine since the function is present with the same type signature in that version too. The side effect is that 

I expect CI will fail, due to linting errors. Running `npm run lint` even on the unmodified 3.x branch is causing a bunch of linting errors on my system. I do not think that fixing pre-existing lint errors was in scope for this pull request. (Weird. CI did not break. Wonder what is up with that)

Note: I had to force push because i needed to rebase this on top of 3.5.3 instead of 3.5.2.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -355,6 +355,150 @@
         "rxjs": "6.3.3"
       }
     },
+    "@angular/cli": {
+      "version": "7.3.10",
+      "resolved": "https://registry.npmjs.org/@angular/cli/-/cli-7.3.10.tgz",
+      "integrity": "sha512-ikSwYa3Pc8GkfyoQ+bzblxyuPnKzNYD6SRVbO4CXUbTJB+NmOCNEV3LTwwMANVBp4Z6fBMdUZccJdFbT4FbG2Q==",
+      "dev": true,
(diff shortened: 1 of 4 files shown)


KevinCathcart:
The only meaningful differences from the 4.x branch are: not using `?.` for safe navigation, as the used typescript version did not support it.

The only other differences are:

- a few extra `// eslint-disable-next-line @typescript-eslint/no-use-before-define`
- lack of `: any` type annotation on the exported function, since the config object lacked such an annotation in 3.x. (I.e. for consistency between both function parameters).
- no change to the tests, since the test that needed to be updated in 4.x does not exist in 3.x

KevinCathcart:
Yeah. I'm really confused about CI passing when linting fails on my
machine, but whatever, not too worried about it.

Issue #206, opened 2020-06-01
[BUG] NgZone reference gets overridden when dependencies are shared


arturovt:
After I started investigating _sharing dependencies approach_ in Angular I have noticed that `ngZone` gets overridden by the last mounted application here:

https://github.com/single-spa/single-spa-angular/blob/389a16cb186ed5fd975199ff2e4670084c47a84f/src/src/extra-providers.ts#L73-L75

So if we've got 2 active applications:
* both of them will call `setNgZone(ngZone)`
* since `SingleSpaPlatformLocation` is a part of the `PLATFORM_INJECTOR` then the second application will override `ngZone` with its own instance
* the first application will not be notified about `popstate` events because here:
https://github.com/single-spa/single-spa-angular/blob/389a16cb186ed5fd975199ff2e4670084c47a84f/src/src/extra-providers.ts#L63
`ngZone` is coming from 2nd application.

The solution is to keep references to all `ngZone`'s:

```ts
export class SingleSpaPlatformLocation {
  private ngZones: NgZone[] = [];

  setNgZone(ngZone: NgZone): void {
    this.ngZones.push(ngZone);
  }

  onPopState() {
    // ...
    for (const ngZone of this.ngZones) {
      ngZone.run(() => fn(event));
    }
  }

  destroy(ngZone: NgZone): void {
    // ...
    const ngZoneIndex = this.ngZones.indexOf(ngZone);
    this.ngZones.splice(ngZoneIndex, 1);
  }
}
```

We also have to take into consideration that part of code:

https://github.com/single-spa/single-spa-angular/blob/389a16cb186ed5fd975199ff2e4670084c47a84f/src/src/extra-providers.ts#L33

Since it will clear all `popstate` event listeners even if one of the applications is still active.

joeldenning:
Good catch - yes let's switch to an array of ngZones.

filoxo:
Closing since this was resolved in #262 and released in 4.4.0.

PR #262, opened 2020-08-18
fix: use `Zone.wrap` except of NgZone instance since it is overriden when deps are shared

arturovt:


arturovt:
> Can you explain this change? What do you mean by "when deps are shared?" And how are things overwritten?

#206

arturovt:
The problem I've encountered when trying to fix that issue is removing application specific `popstate` listeners. E.g.:

* shop application adds `popstate` listener
* navbar application adds `popstate` listener

When shop is destroyed we'd want to remove only its `popstate` listener.

The `SingleSpaPlatformLocation` is reused between 2 applications and I can determine the current `onPopState` execution context only via `Zone.current`, since I don't know what application invokes this method.

src/src/extra-providers.ts
```diff
@@ -1,36 +1,64 @@
...
+  // by application `name`.
+  // The reason of that the `onPopState` method is invoked during `bootstrapModule`
+  // and we can't know what application has invoked it. Why should we know the application
+  // that has invoked `onPopState`? When application gets destroyed in a `sharing dependencies mode`
+  // (when there is a single platform per all applications) we want to remove application
+  // specific `popstate` listeners. E.g. if there are 2 applications:
+  // * shop application adds `popstate` listener
+  // * navbar application adds `popstate` listener
+  // When shop application gets destroyed we want to remove only its `popstate` listener.
+  private zoneToOnPopStateListenersMap = new Map<any, OnPopStateListener[]>();
```
joeldenning, line 30: Why can't this map have the zoneIdentifier as key? Surely we can extract the zone identifier from the zone and use that as the key? I've read through your comments here and am still unsure about it.

src/src/extra-providers.ts
```diff
@@ -1,36 +1,64 @@
...
+
+  // This is used only to make `Zone.wrap` happy, since it requires 2 arguments
+  // and the second argument is a unique string which `zone.js` uses for debugging purposes.
+  // We might want to use the application name, but we're not able to get it when `onPopState`
+  // method is called during module bootstrapping.
+  private source = 0;
```
joeldenning, line 36: 👍 nice explanation of this

src/src/single-spa-angular.ts
```diff
@@ -119,21 +119,19 @@ async function mount(options: SingleSpaAngularOptions, props: any): Promise<NgMo
...
+    const zoneIdentifier: string = bootstrappedOptions.zoneIdentifier!;
 
     // `NgZone` can be enabled but routing may not be used thus `getSingleSpaExtraProviders()`
     // function was not called.
     if (singleSpaPlatformLocation !== null) {
-      singleSpaPlatformLocation.setNgZone(ngZone);
       // Cleanup resources, especially remove event listeners thus they will not be added
       // twice when application gets bootstrapped the second time.
-      module.onDestroy(() => singleSpaPlatformLocation.destroy());
+      module.onDestroy(() => singleSpaPlatformLocation.destroyApplication(zoneIdentifier));
```
joeldenning, line 130: 👍 i like this approach

src/src/extra-providers.ts
arturovt: Yep, it's not accessible yet because this `onPopState` method is called during `await bootstrapFunction()` and before we set our custom property.

PR #293, opened 2020-11-03
feat: support Nrwl Nx workspace

arturovt:


joeldenning (approved):
There are a lot of changes here, but this looks good to me. Excited for an easier workflow with nx!

PR #295, opened 2020-11-03
build: upgrade webpack to version 5

arturovt:
Adding Webpack to 5 version to test module federation feature.

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,6 +119,7 @@
     "tsickle": "^0.39.1",
     "tslib": "^2.0.1",
     "typescript": "3.9.7",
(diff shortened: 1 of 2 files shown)


joeldenning:
I don't think adding webpack 5 to the main package.json will change which version of webpack Angular CLI will use, since Angular CLI has a dependency on webpack 4. The people doing Angular + module federation have custom webpack configs instead of using Angular CLI

arturovt:
Yeah, I'm testing the custom config right now.

arturovt:
Closing as discussed in Slack will focus on the import maps investigation.

PR #298, opened 2020-11-06
feat: implement `single-spa-angular/parcel` library to render parcels

arturovt:


package.json
```diff
@@ -142,9 +151,6 @@
     }
   },
   "lint-staged": {
-    "*.{ts,html,scss,md,yml}": [
-      "prettier --write",
-      "git add"
```
joeldenning, line 147: Does lint-staged automatically stage git files now?

libs/single-spa-angular/parcel/src/parcel.component.ts
```diff
@@ -0,0 +1,128 @@
...
+      }
+    });
+  }
+
+  ngOnInit(): void {
+    this.scheduleTask(Action.Mount, () => {
+      if (this.mountParcel === null) {
+        throw new Error(`
+				  <parcel> was not passed a [mountParcel] binding.
+				  If you are using <parcel> within a module that is not a single-spa application, you will need to import mountRootParcel from single-spa and pass it into <parcel> as a [mountParcel] binding
```
joeldenning, line 53: One thing to note here, is that there is `mountRootParcel` and `mountParcel`. The `mountParcel` function is one that is provided as a prop to all single-spa applications. The two functions are identical, except that `mountParcel` is bound to the current application, and will automatically unmount the parcel when the application is unmounted.

In React, we automatically use the `mountParcel` function inside of single-spa-react by using React context to do dependency injection.

Let's update the languaging here to reflect this:

```suggestion
				  If you are using <parcel> within a module that is not a single-spa application, you will need to either (1) import mountRootParcel from single-spa or (2) use the `mountParcel` prop provided to single-spa applications. One of those two should be passed into <parcel> as a [mountParcel] binding
```

libs/single-spa-angular/parcel/src/parcel.component.ts
```diff
@@ -0,0 +1,128 @@
...
+				`);
+      }
+
+      this.wrapper = document.createElement(this.wrapWith);
+
+      if (this.appendTo !== null) {
+        this.appendTo.appendChild(this.wrapper);
+      } else {
+        this.host.nativeElement.appendChild(this.wrapper);
```
joeldenning, line 62: Is `host.nativeElement` the parent component's wrapper dom element?

apps/parcel/src/app/ReactWidget/ReactWidget.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import singleSpaReact from 'single-spa-react';
+
+const ReactLogo = React.lazy(() => import('./ReactLogo'));
+
+class ReactWidget extends React.Component {
```
joeldenning, line 7: The Angular dev in you is coming out here, in choosing to use a react class component instead of function component 😄  (this is a joke - using class components is completely fine)

cypress/plugins/index.js
```diff
@@ -18,4 +18,14 @@
 module.exports = (on, config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
+
+  // https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
+  const isCI = process.env.CI === 'true';
+  console.log(`[isCI = ${isCI}]`);
```
joeldenning, line 24: Should this be removed?

apps/parcel/src/app/app.module.ts
```diff
@@ -0,0 +1,12 @@
+import { BrowserModule } from '@angular/platform-browser';
+import { NgModule } from '@angular/core';
+import { ParcelModule } from 'single-spa-angular/parcel';
```
joeldenning, line 3: Why don't we have to import ParcelComponent here and add it to the declarations block?

cypress/integration/parcel.spec.js
```diff
@@ -0,0 +1,14 @@
...
+describe('Angular parcel', () => {
+  it('should navigate to /parcel and render the React widget and lazy component', () => {
+    cy.visit('/parcel')
+      // GitHub Actions CI is not as fast as the local setup, there can be some network delays,
+      // timeout is used only for this purpose.
+      .get('parcel-root parcel', { timeout: Cypress.env('timeout') })
+      .should('exist')
+      .get('parcel-root parcel img')
+      .invoke('attr', 'alt')
+      .should('eq', 'React logo');
```
joeldenning, line 12: 👍  nice!

package.json
arturovt: Don't really know :) It was whining that some hook uses `git add` :man_shrugging:

joeldenning (approved):
Great work

PR #344, opened 2021-03-01
fix: remove direct workspace access

arturovt:


joeldenning:
What changes to angular schematics are coming in upcoming versions?

arturovt:
Hey, Angular schematics package has an API for working with workspaces and projects w/o accessing properties directly like that:
```js
clientProject.architect!.build!.configurations
```

* I haven't done any functional changes in this PR. I've only applied the Angular's way of dealing with workspaces and projects, it's mentioned here: https://angular.io/guide/schematics-for-libraries#get-the-project-configuration.
* I moved all rules from the `chain()` function into `rules` folder since the `index.ts` file becomes messy and it's a bit complicated to read it

---

I should've actually renamed PR to `fix: remove direct workspace access` since it's not `feat`.

joeldenning (approved):
Being honest I didn't really look this over in much detail, as I don't know this code very well and it was a lot of changed files. But I'm going to approve and merge now so that my lack of knowledge doesn't delay progress on the project.

Issue #370, opened 2021-06-08
Sharing Angular with Ivy throws: "Class constructor BrowserPlatformLocation cannot be invoked without 'new'"


OriolInvernonLlaneza:
While preparing for the upgrade to `Angular 12`, I tried to change our shared `Angular 11` setup that is working with `ViewEngine` to `Ivy` activated like in the example at the [angular-microfrontends repo](https://github.com/angular-microfrontends). 
However, if I keep sharing `single-spa-angular` in order to workaround the problem discussed at #288 ([see](https://github.com/single-spa/single-spa-angular/issues/288#issuecomment-720380317)), I receive the following error:
```
Uncaught TypeError: Class constructor BrowserPlatformLocation cannot be invoked without 'new'
    at new SingleSpaPlatformLocation (extra-providers.ts:13)
```
If I, like the angular-microfrontends example, stop sharing `single-spa-angular` I get the same error as in #288:
```
Uncaught Error: 
      single-spa-angular: could not retrieve extra providers from the platform injector. Did you call platformBrowserDynamic(getSingleSpaExtraProviders()).bootstrapModule()
```

### Demonstration
I'll try to make a demo later. 
Meanwhile, my config looks like this:

Angular on the package.json:
```
  "dependencies": {
    "@angular/common": "~11.1.1",
    "@angular/compiler": "~11.1.1",
    "@angular/core": "~11.1.1",
    "@angular/forms": "~11.1.1",
    "@angular/platform-browser": "~11.1.1",
    "@angular/platform-browser-dynamic": "~11.1.1",
    "@angular/router": "~11.1.1",
    "angular-i18next": "~10.2.0",
    "i18next": "^20.3.1",
    "rxjs": "^6.6.0",
    "single-spa": "~5.8.0",
    "single-spa-angular": "~4.9.0",
    "zone.js": "~0.11.4"
  },
```
The webpack config:
```
module.exports = (angularWebpackConfig, options) => {
  const singleSpaWebpackConfig = singleSpaAngularWebpack(angularWebpackConfig, options);

  singleSpaWebpackConfig.externals.push(
    /^@apps\/*/,
    /^single-spa$/,
    /^single-spa-angular$/,
    /^single-spa-angular\/internals$/,
    /^rxjs$/,
    /^rxjs\/operators$/,
    /^@angular\/core$/,
    /^@angular\/common$/,
    /^@angular\/router$/,
    /^@angular\/platform-browser$/,
    /^zone\.js$/,
[...]

PR #374, opened 2021-06-25
fix: conversion of UMD bundle to ES5 avoided using patch for ng-packagr

ajaykrishnan33:
## PR Checklist

## PR Type

```
[x] Bugfix
[x] Build related changes
```

Issue Number: https://github.com/single-spa/single-spa-angular/issues/370

## What is the new behavior?

The UMD bundles generated for single-spa-angular will be in ES2015 syntax instead of being in ES5 syntax.

```
[x] No
```

### DIFF
--- a/libs/single-spa-angular/package.json
+++ b/libs/single-spa-angular/package.json
@@ -1,7 +1,7 @@
 {
   "$schema": "../node_modules/ng-packagr/ng-package.schema.json",
   "name": "single-spa-angular",
-  "version": "4.9.2",
+  "version": "4.9.3",
   "description": "Helpers for building single-spa applications which use Angular 2",
   "schematics": "./schematics/schematics.json",
   "repository": {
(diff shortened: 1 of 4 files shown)


libs/single-spa-angular/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "../node_modules/ng-packagr/ng-package.schema.json",
   "name": "single-spa-angular",
-  "version": "4.9.2",
+  "version": "4.9.3",
```
joeldenning, line 4: Please revert, as changing the version is done during publish.

patches/ng-packagr+11.2.4.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/ng-packagr/lib/ng-package/entry-point/write-bundles.transform.js b/node_modules/ng-packagr/lib/ng-package/entry-point/write-bundles.transform.js
+index ecb19da..2cd9df4 100644
+--- a/node_modules/ng-packagr/lib/ng-package/entry-point/write-bundles.transform.js
++++ b/node_modules/ng-packagr/lib/ng-package/entry-point/write-bundles.transform.js
+@@ -54,7 +54,7 @@ exports.writeBundlesTransform = transform_1.transformFromPromise(async (graph) =
+             format: 'umd',
+             dest: umd,
+             cache: cache.rollupUMDCache,
+-            transform: downlevel_plugin_1.downlevelCodeWithTsc,
++            //transform: downlevel_plugin_1.downlevelCodeWithTsc,
```
joeldenning, line 10: This is a last resort. I'm not opposed to it if we can't solve this another way, but would prefer exploring other options first. See https://github.com/single-spa/single-spa-angular/issues/370#issuecomment-875290631 where i discuss an alternative

arturovt:
I'll close the PR since UMD generation is no longer enabled starting from Angular 13.

Issue #379, opened 2021-07-14
single-spa latests and angular 12, fails to compile: Invalid parser object...


pgarzina:
### Demonstration

We have updated one of our microfrontend apps to single-spa latest and angular 12, and it stoped compiling.

the issue comes down to our ` environment.ts` file:

```
declare var require: any;

export const environment = {
  production: false,
  API_URL: '',
  API_STATIC_URL: '',
  ...
  app_version: require('../../package.json').version,
  library_version: require('../../node_modules/@cognism/app-library/package.json').version,
};
```

We read both our current app and our lib version from their package.json files. And that worked fine until updating both the single spa single spa dependencies to latest and angular to 12.

We also created a blank angular 12 project, and reading from package.json works just fine. After adding single-spa `ng add single-spa-angular`, it does not compile anymore and throws an error, which seems like the issue lies somewhere in single-spa or some of its dependencies.

We tried importing with `import` along with `allowSyntheticDefaultImports` and `resolveJsonModule` but that fails as well.

During the update we moved from
`"single-spa": "^5.9.1"` -> `"single-spa": "^5.9.3"`,
`"single-spa-angular": "^4.9.1"` -> `"single-spa-angular": "^5.0.1"`,
`"@angular-builders/custom-webpack": "^10"` -> `"@angular-builders/custom-webpack": "^12.1.0"`,
`"webpack-merge": "^5.7.3"` -> `"webpack-merge": "^5.8.0"`

Feel free to ask any additional info if needed.
Thanks

### Expected Behavior

The application should compile.

### Actual Behavior

The application does not compile and throws an error

```
./src/environments/environment.ts:4:0-41 - Error: Module not found: ValidationError: Invalid parser object. Json Modules Plugin has been initialized using a parser object that does not match the API schema.
 - parser has an unknown property 'system'. These properties are valid:
   object { parse? }

```

pgarzina:
the compiler throws an error in this case as well
```
import { version as libVersion } from '../../node_modules/@cognism/app-library/package.json';
import { version as appVersion } from '../../package.json';

@Component({
  selector: 'app-search-root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css'],
})
export class AppComponent {
  constructor(private renderer: Renderer2) {
    console.log(`appVersion `, appVersion);
    console.log(`libVersion `, libVersion);
  }
```

joeldenning:
[...]

PR #380, opened 2021-07-23
fix: invalid parser object (issue #379)

OriolInvernonLlaneza:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[X] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The Webpack rule `"parser": {"system": false }` has been removed from Webpack 5. This causes an error in compile time due to  @angular-builders/custom-webpack v12 using Webpack 5.

Issue Number: #379 

## What is the new behavior?
The Angular 12 microfrontend is correctly compiled.

```
[ ] Yes
[X] No
```
At least I don't think it does but maybe I'm wrong here.

## Other information
All of Jest and Cypress tests passed on local.

### DIFF
--- a/libs/single-spa-angular/package.json
+++ b/libs/single-spa-angular/package.json
@@ -1,7 +1,7 @@
 {
   "$schema": "../node_modules/ng-packagr/ng-package.schema.json",
   "name": "single-spa-angular",
-  "version": "5.0.2",
+  "version": "5.0.1",
   "description": "Helpers for building single-spa applications which use Angular 2",
   "schematics": "./schematics/schematics.json",
   "repository": {
--- a/libs/single-spa-angular/package.json
+++ b/libs/single-spa-angular/package.json
@@ -1,7 +1,7 @@
 {
   "$schema": "../node_modules/ng-packagr/ng-package.schema.json",
   "name": "single-spa-angular",
-  "version": "5.0.1",
+  "version": "5.0.2",
   "description": "Helpers for building single-spa applications which use Angular 2",
   "schematics": "./schematics/schematics.json",
   "repository": {
--- a/libs/single-spa-angular/webpack/index.ts
+++ b/libs/single-spa-angular/webpack/index.ts
@@ -43,9 +43,6 @@ export default (config: any, options?: Options, extraOptions?: DefaultExtraOptio
     module: {
       rules: [
         {
-          /*parser: {
-            system: false,
-          },*/
         },
       ],
     },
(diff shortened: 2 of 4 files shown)


libs/single-spa-angular/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "../node_modules/ng-packagr/ng-package.schema.json",
   "name": "single-spa-angular",
-  "version": "5.0.1",
+  "version": "5.0.2",
```
joeldenning, line 4: Please revert - version numbers are bumped during publish, not in each PR

libs/single-spa-angular/webpack/index.ts
```diff
@@ -43,9 +43,9 @@ export default (config: any, options?: Options, extraOptions?: DefaultExtraOptio
     module: {
       rules: [
         {
-          parser: {
+          /*parser: {
```
joeldenning, line 46: Does angular 12 use webpack 5? The parser system false is required in webpack 4, but not 5.

Also, can you delete this code rather than comment it out?

libs/single-spa-angular/package.json
OriolInvernonLlaneza: Oh sorry. I updated it.

libs/single-spa-angular/webpack/index.ts
OriolInvernonLlaneza: At least when I tried to get it working in fresh demo projects or our MFEs I always got that error, even when explicitly declaring Webpack 4 in the dependencies. I looked at yarn.lock and saw that @angular-builders/custom-webpack v12 forced Webpack 5 on its dependencies. So I assumed this was the error, came here looking for issues and found #379.

I'm not at work today so I can't show you the yarn.lock right now.

OriolInvernonLlaneza: yarn.lock:
https://github.com/survivejs/webpack-merge/blob/c658e1904a1ebc440fdda6fc88fca1332fb02ea9/package.json#L29
image
@angular-devkit/build-angular also has v5 as dependency.

Error:
image

package.json:
image

joeldenning: The first screenshot shows the webpack-merge version for `@angular/custom-webpack`, not the webpack version. Angular ignores any webpack versions in the package.json. Instead, it uses the version specified by `@angular-devkit/build-angular`.

I tested it out and confirmed that it does use webpack 5.

joeldenning (approved):

joeldenning:
Released in https://github.com/single-spa/single-spa-angular/releases/tag/v5.0.2

libs/single-spa-angular/webpack/index.ts
OriolInvernonLlaneza: I didn't knew that. Glad it could be merged. Thank you, Joel!

PR #428, opened 2022-04-01
feat: add `excludeAngularDependencies` option to exclude shared dependencies

arturovt:


libs/single-spa-angular/webpack/index.ts
```diff
@@ -54,6 +55,51 @@ export default (config: any, options?: Options, extraOptions?: DefaultExtraOptio
     devtool: resolveDevtool(options),
   };
 
+  if (options?.customWebpackConfig?.excludeAngularDependencies) {
+    // Note: we need to specify packages explicitly w/o wildcards (e.g. /@angular\/*/). The developer
+    // may have an alias that also starts with `@angular/`, this may lead to unexpected behavior.
```
filoxo, line 60: This is awesome! I like moving this into the helper itself. Question: will this also have documentation that points to these libs? I often help people with stuff on Slack and it would be great to be able to point them to docs that show them where to get these shared libs that they might need to add. Probably in this section: https://single-spa.js.org/docs/ecosystem-angular/#shared-angular

arturovt: Hey, I already cloned docs repo to update the documentation and add a guide for sharing dependencies since I was finally able to make it 😄

filoxo: Awesome, thank you!

PR #465, opened 2022-12-29
feat: upgrade to Angular 15

arturovt:


ALP011216:
It seems that the type definitions of the `single-spa-angular` function are no longer included in the v8 package.
https://www.jsdelivr.com/package/npm/single-spa-angular?path=src

arturovt:
I see, this is an issue in ng-packagr, I will find time to fix it and publish 8.0.1

PR #480, opened 2023-06-22
feat: upgrade to Angular 16

arturovt:


bumalai:
@arturovt 
Hi! Would you look at this MR https://github.com/single-spa/single-spa-angular/pull/488
I think it will fix the tests.

bumalai:
@arturovt ?

PR #513, opened 2024-04-19
fix: `assertNotInAngularZone` should act as a noop

arturovt:


victorlevasseur:
Thank you very much, it fixes the infinite errors we were getting in the console in dev mode, preventing us from testing ours apps with Angular 17+

PR #541, opened 2025-04-13
Fix types

jolyndenning:


src/single-spa-angular.ts
```diff
@@ -1,27 +1,14 @@
-import { registerApplication, mountRootParcel } from "single-spa";
+import { Application, AppProps, ParcelConfigObject } from "single-spa";
 import { chooseDomElementGetter } from "dom-element-getter-helpers";
 import { ApplicationConfig, Type } from "@angular/core";
 
-type CustomProps = Parameters<typeof mountRootParcel>[1];
```
jolyndenning, line 5: This types no longer needed now that single-spa@7 exports the needed types

PR #542, opened 2025-04-13
Fix types again

jolyndenning:


src/single-spa-angular.ts
```diff
@@ -1,10 +1,10 @@
-import { Application, AppProps, ParcelConfigObject } from "single-spa";
+import { AppProps, LifeCycles } from "single-spa";
 import { chooseDomElementGetter } from "dom-element-getter-helpers";
 import { ApplicationConfig, Type } from "@angular/core";
 
 export function singleSpaAngular<Props>(
   opts: SingleSpaAngularOpts<Props>,
-): Application<Props> | ParcelConfigObject<Props> {
+): LifeCycles<Props> {
```
jolyndenning, line 7: Application type is a union of lifecycles or a loading function for the lifecycles. singleSpaAngular() never returns a loading function

PR #575, opened 2026-06-10
fix: accept thenable bootstrapFunction return values (#574)

### LINKED ISSUE #574: Error: the options.bootstrapFunction must return a promise
Hello, I am getting this error since we upgraded to Angular 21:

```
Uncaught Error: application 'smc-mf-content' died in status SKIP_BECAUSE_BROKEN: single-spa-angular: the options.bootstrapFunction must return a promise, but instead returned a 'object' that is not a Promise
    at mount (single-spa-community-angular.mjs:206:15)
    at lifecycle.helpers.js:28:29
```

It seems to come from [this line](https://github.com/single-spa/single-spa-angular/blob/b713ce619861aade1ee7d11d5e00ab67c3be5894/libs/single-spa-angular/src/single-spa-angular.ts#L102):
```ts
if (!(bootstrapPromise instanceof Promise)) {
```

We did not have it before, so I am thinking that probably something internal in Angular changed that impacts the bundling. We use the Angular `application` builder (esbuild). Or maybe it's because we "externalize" `single-spa` via our importmap? (but we were doing that already before...)

In any case, I took a look at what single-spa itself is doing internally and I believe [that implementation](https://github.com/single-spa/single-spa/blob/297483c11265050542d7ef107d14c44936c51883/src/lifecycles/lifecycle.helpers.js#L29) is a bit more robust:
```ts
export function smellsLikeAPromise(promise) {
  return (
    promise &&
    typeof promise.then === "function" &&
    typeof promise.catch === "function"
  );
}
```

I have tried patching single-spa-angular with this `smellsLikeAPromise` approach (instead of `instanceof Promise`) and that works like a charm for us.

Would you accept a PR for this `smellsLikeAPromise` approach on this repo?

kfrederix:
## Summary
- Replace `instanceof Promise` with duck-typed promise detection when validating `bootstrapFunction` return values
- Add `smellsLikeAPromise` helper aligned with single-spa's own promise detection
- Fixes Angular 21 + esbuild/importmap setups where bootstrap promises come from a different Promise constructor

Fixes #574

## Test plan
- [x] Added unit tests for `smellsLikeAPromise` (native promises, foreign thenables, invalid values)
- [x] `yarn test`
- [x] Verify affected microfrontend mounts successfully after upgrade

@arturovt please let me know if you have any concerns about this?


libs/single-spa-angular/src/single-spa-angular.ts
```diff
@@ -3,6 +3,7 @@ import type { LifeCycles } from 'single-spa';
 import { getContainerElementAndSetTemplate } from 'single-spa-angular/internals';
 
 import { SingleSpaPlatformLocation } from './platform-providers';
+import { smellsLikeAPromise } from './smells-like-a-promise';
```
arturovt:
```suggestion
import { ɵisPromise } from '@angular/core';
```

libs/single-spa-angular/src/single-spa-angular.ts
```diff
@@ -99,7 +100,7 @@ async function mount(
 
   const bootstrapPromise = options.bootstrapFunction(props);
 
-  if (!(bootstrapPromise instanceof Promise)) {
+  if (!smellsLikeAPromise(bootstrapPromise)) {
```
arturovt:
```suggestion
  if (!ɵisPromise(bootstrapPromise)) {
```

arturovt (changes requested):

libs/single-spa-angular/src/single-spa-angular.ts
kfrederix: updated

libs/single-spa-angular/src/single-spa-angular.ts
kfrederix: updated

arturovt (approved):

arturovt:
I'll merge this afternoon and publish.

